PR #11, opened 2019-12-02
add read_acqknowledge and etc

Tam-Pham:
# Description

added new functions to data 
add bioread into deps

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -6,8 +6,10 @@ Core team
 ----------------
 
 * Dominique Makowski <dom.makowski@gmail.com>
+* Tam Pham <phamttam17@gmail.com>
 
 Contributors
 ------------
 
-None yet. Why not be the first?
+* Hung Pham <hungpham2511@gmail.com>
+
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -5,11 +5,11 @@ Credits
 Core team
 ----------------
 
-* Dominique Makowski <dom.makowski@gmail.com>
-* Tam Pham <phamttam17@gmail.com>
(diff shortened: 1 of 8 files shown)


DominiqueMakowski:
@Tam-Pham could you add yourself in https://github.com/neuropsychology/NeuroKit/blob/master/AUTHORS.rst in core team and @hungpham2511 to contributors?

Thanks @hungpham2511 😉

hungpham2511:
@DominiqueMakowski You are welcome! Nice, thanks!

DominiqueMakowski:
Now we'd need to add some tests for the functions, which in turn requires to find a neat and efficient way to store data, both to use it in tests and to have some reproducible documentation. 

In neurokit 1 the data was stored as csv in an example folder (https://github.com/neuropsychology/NeuroKit.py/tree/master/examples/Bio) (used in [the documentation](https://github.com/neuropsychology/NeuroKit.py/blob/master/docs/tutorials/Bio.rst#preparation)) and *also* in the tests folder (https://github.com/neuropsychology/NeuroKit.py/tree/master/tests/data). However, the latter seems to not have been further used in the tests (guess I didn't manage to do that).

@hungpham2511 do you by any chance have an idea of the most efficient way to store data so that it could be used in the tests and documentation, but at the same time that would *not* be shipped with the package (to avoid long downloading times)?

hungpham2511:
@DominiqueMakowski I am sure you can configure the `setup.py` in such a way that tests/data are not included in the built distribution to be pushed to PyPI. To avoid downloading, say during, git pull, probably you can put the data to a separate repository and only download when necessary.

DominiqueMakowski:
Thanks, let's merge this for now, we'll address that later ☺️

PR #29, opened 2019-12-09
Proposal for Breathing Module

JohnDoenut:
# Description

Dear all,
I would like to add a breathing module and made the following changes.

# Proposed Changes

I added a `rsp` module that contains a `rsp_process()` function. You find a short demo and description of the `rsp` module in scripts/respiration.ipynb. I think we should spend some time discussing the `Rsp` object in `rsp_process()`, because I guess that we want to have a uniform API across all modalities and `Rsp` could be a template if we agree on its implementation.

The `rsp` module requires two helper functions:
* To be able to filter the breathing signal, I added a `signal_filter()` function to the `signal` module. It contains a wrapper for the Butterworth IIR filter implemented in [scipy.signal](https://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.signal.butter.html). Maybe it's useful for other modules too. You find a short demo of `signal_filter()` in scripts/butterworth_filter.ipynb. Maybe we want to test it more formally (I wouldn't know how atm)?
* To get instantaneous statistics (i.e., statistics over time), I added an `interpolate()` function to the `stats` module. Note that this can also be used for PPG and ECG (basically any signal in which statistics are derived from differences between extrema).

Lastly, I added a breathing test signal to /data. I guess that eventually we can remove it once we can simulate breathing signals. So consider it a temporary solution to have some data for the respiration.ipynb notebook.


DominiqueMakowski:
@JohnDoenut Awesome!! Will merge and take a look ☺️  Do you have on the top of your head some lines to put in [tests](https://github.com/neuropsychology/NeuroKit/tree/master/tests) so we could get that test coverage up?

JohnDoenut:
@DominiqueMakowski, I'll write some tests asap. For `rsp_process()` the test would be based on /data/testsignal_rsp.txt though. Meaning we'll have to update the test once we run it with simulated data. But that should only be a small adaptation then I guess.

PR #30, opened 2019-12-10
Dev

DominiqueMakowski:
- [ ]  rsp_process


DominiqueMakowski:
@JohnDoenut great work on that rsp processing! I have made some modifications to the arg names for consistency (e.g. `sfreq` -> `sampling_rate`). Also, I started 'outsourcing' the function, as I think some of the things that happen inside could be done outside (and  reused, for instance the preprocessing, to find peaks, to compute rate etc.). 

Thus, I am adding some functions to the signal submodule (for instance, signal_findpeaks - do you think it could replace the current chunk of extremas locating?). 

Instead of a unique class for `rsp_process`, I think it could be decomposed on several functions, at minimum:
- `rsp_prepare()`: filter, detrend and all
- `rsp_process()`: would call the preparation and then compute the stats (breathing rate etc.)

what do you think? I've started writing rsp_prepare, but not sure if everything is correct there! I hope you can take a look ☺️

JohnDoenut:
@DominiqueMakowski , 

+1 for renaming `sfreq`. 

I wouldn't want to re-purpose the extrema detection used in `Rsp` though. It's very ideosyncratic to this [algorithm](https://www.biorxiv.org/content/10.1101/270348v1.full) and requires a zero-centered signal (hence the detrending). 

I don't see the advantage of pulling `rsp_prepare()` out of the `Rsp` class. Is there any other use for `rsp_process()` outside of `Rsp`? I would only re-use functions from `Rsp` if they are generic enough to be useful in other contexts. I feel like this would only be the case for the filtering and interpolation.

I guess the discussion ultimately comes down to how we want the API to look like. I prefer giving the user some more granularity in their choice as to what they want to get out of the signal. I.e., they instantiate the breathing object (`Rsp`) and can then choose to return only those attributes that they're interested in. I.e., a user might only want to look at the signal or only have the extrema returned. This would be in contrast to a `rsp_process()` function that returns all attributes at once in a nested dict like it was the case in [Nk1](https://github.com/neuropsychology/NeuroKit.py/blob/master/neurokit/bio/bio_rsp.py). The latter feels less intuitive to me.

Happy to hear your thoughts on this. The above is my preference regarding the API but I'm absolutely open to alternatives :)

DominiqueMakowski:
@JohnDoenut I have given it a try to open the Rsp class, with 3 functions, rsp_prepare, rsp_findpeaks and rsp_rate. The names were chosen so that the pattern can be used in other signals, for instance, ecg there would also be ecg_prepare, ecg_findpeaks and ecg_rate, in order to keep the api to a minimum by maximizing consistency across signals (while allowing for some flexibility and specificity). Then, we have the `rsp_process` metafunction that integrates all of the above (arguably syntactic sugar for new users that just want to have a rough idea of what they can obtain).

Currently, some of the functions return two things, but in the same pattern, a dataframe (containing signals) and a info dict containing some specific information that can be of various sizes and shapes (for instance, the indices of peaks, troughs etc.). 

Just lacks the rsp_plot() function, I'll add it now.

I added examples in the docstrings that should work. How does that implementation feel?

JohnDoenut:
> The names were chosen so that the pattern can be used in other signals, for instance, ecg there would also be ecg_prepare, ecg_findpeaks and ecg_rate, in order to keep the api to a minimum by maximizing consistency across signals (while allowing for some flexibility and specificity).

I like that apporach very much! Althoug I find that xxx_rate could be a misnomer since in this case it returns amplitude as well. So maybe xxx_stats woud be more descriptive and accomodating of other modalities? Also maybe xxx_clean instead of xxx_prepare? I guess it's easy to get lost in that semantic rabbit-hole, but it's about exposing the concepts within a biosignal worflow as intuitively as possible. I.e., a typical workflow would be: clean/prepare (xxx_clean) --> identify interesting points in the signal (xxx_findpeaks) --> retrieve information from these interesing points (xxx_stats) --> visualize all (or parts) of it (xxx_plot).

DominiqueMakowski:
Haha I am totally in line with the importance of the right names (in some other project, (easystats for R), we have like 60posts+ threads debating whether to name something `model_parameters` or `parameters_model` 😅) So please continue sharing your thoughts!! We want to really think things through for neurokit2 ☺️ 

I agree, `*_clean()` is better than prepare, will change it. As for `_rate`, I agree that it's not perfect since it returns amplitude as well... but `stats` feels awkward too, as I feel like this is not what people would expect from it (likely they'd expect some real stats and tests and whatnot, something we might add later on [there will be at least a `_eventrelated()` function or something like this to compute changes in relationship with events, so from there we might need a `_stats()` function]). 

At the same time, since amplitudes are also related to the rate (its changes are of the same "rate" as rate), it doesn't bother me *that* much *for now* to have it under _rate (especially since people will be googling "get respiration rate python" and they would expect a `rsp_rate` to do that, even though it does more than that). Let's keep this in mind in case we stumble on a better name!

DominiqueMakowski:
Before merging:

- [ ] Decide on whether to keep or remove Rsp class for now
- [ ] Documentation for `rsp_findpeaks`
- [ ] Documentation for `rsp_rate`
- [ ] Documentation for `rsp_process`
- [ ] Testing (not a priority, can be done later)

@JohnDoenut would you have a bit of text for the docstrings of the functions, to describe a bit what/why it does, as well as some refs?

JohnDoenut:
@DominiqueMakowski , I added some docstrings in #33. As far as I'm concerned we can remove `Rsp`.

PR #52, opened 2019-12-20
Dev

DominiqueMakowski:


DominiqueMakowski:
@zen-juen nice tests, but the last lines don't work ☺️ , if you look at the travis output and why it fails (https://travis-ci.org/neuropsychology/NeuroKit/jobs/628034624#L755), you can see that it says that `module 'neurokit2' has no attribute '_eda_simulate_canonical'`. This is because internal functions (those starting with `_*`) are not usually not exported, i.e., not available from outside the package (there are just used within the package, but not meant to be called directly by end-users). Which is another challenge for testing, as we can only test exported functions 🙉 nice try tho!

PR #57, opened 2019-12-26
dev

DominiqueMakowski:


DominiqueMakowski:
@CShoel I moved the updated entro-py code to [entropy_fuzzy](https://github.com/neuropsychology/NeuroKit/blob/dev/neurokit2/complexity/entropy_fuzzy.py#L9) as it is only used for that purpose now. However, although fuzzy entropy is not in nolds, would you guess that the implementation is okay? The only changes from regular sample entropy seem to be:

https://github.com/neuropsychology/NeuroKit/blob/bf147e664677104e1e89e487797339f329858760/neurokit2/complexity/entropy_fuzzy.py#L25-L26

and 

https://github.com/neuropsychology/NeuroKit/blob/bf147e664677104e1e89e487797339f329858760/neurokit2/complexity/entropy_fuzzy.py#L38-L42

CSchoel:
I had a quick look at the definition of FuzzyEn in [Chen et al. (2007)](https://doi.org/10.1109/TNSRE.2007.897025). I agree that the changes you mentioned should be everything that is required. I am not familiar with FuzzyEn, so I cannot give any test cases, but the implementation seems fine.

PR #60, opened 2019-12-29
dev

DominiqueMakowski:


DominiqueMakowski:
Thanks detective @CSchoel, I also made the changes in the tests of neurokit2, where the packages that are hard to import (pyeeg, entro-py and entropy) are (manually) duplicated ☺️ 

do you know by any chance if this fix also applies to approximate entropy, as the base code appears as quite similar:

https://github.com/neuropsychology/NeuroKit/blob/50c47a92d772f323adc494e02e522ef7e0f5ce8e/tests/tests_complexity.py#L102

https://github.com/neuropsychology/NeuroKit/blob/50c47a92d772f323adc494e02e522ef7e0f5ce8e/tests/tests_complexity.py#L116

CSchoel:
> do you know by any chance if this fix also applies to approximate entropy, as the base code appears as quite similar:

I am not familiar with ApEn, but from a quick study of the [Richman and Moorman paper](https://www.physiology.org/doi/pdf/10.1152/ajpheart.2000.278.6.H2039) (end of left column on page H2040) it seems that this is one of the biases that is present in ApEn, but removed in SampEn. Therefore I think the code for the approximate entropy should be left unchanged.

PR #61, opened 2019-12-29
translate ecgsyn from matlab to python

duylp:
# Description

This PR aims at adding a pythonic version of ecgsyn


DominiqueMakowski:
🎉

Issue #31, opened 2019-12-11
Resources: ECG processing and features


DominiqueMakowski:
Collection of resources to help us getting the ECG aspect right, feel free to add more:

-----

## R-peak detection & Segmentation

- See #77 

## Artifacts

- [x] [Lipponen & Tarvainen, 2019](https://www.tandfonline.com/doi/abs/10.1080/03091902.2019.1640306)
### General
- [Guidelines for Reporting Articles on Psychiatry and Heart rate variability](https://www.nature.com/articles/tp201673)
- [Slides by @michagaebler (2020) on HRV](https://osf.io/4xdmv/)

### Complexity 

- [Entropy analysis of heart rate variability and its application to recognize major depressive disorder](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6597986/)
- [Entropy Analysis of Short-Term Heartbeat Interval Time](https://www.mdpi.com/1099-4300/19/10/568)

## Other

- [ ] [ECG-RSP phase-amplitude coupling](https://github.com/jusjusjus/pyedr/blob/master/pyedr/coupling.py)

JohnDoenut:
To get us started with ECG processing I'm happy to add an [ECG peak detection](https://github.com/JohnDoenut/biopeaks/blob/master/biopeaks/ecg.py) algorithm to the `ecg` module. I recently benchmarked the algorithm on the [Glasgow University Database (GUDB)](http://researchdata.gla.ac.uk/716/) (see [5. Changelog](https://github.com/JohnDoenut/biopeaks) for details). Then the users could already get heart rate and -period out of their ECG data and we would have a basis for building HRV analyses. If this sounds like a good idea I'll get working on a PR soonish in the new year :) 

On a related note, here's a nice [overview](https://www.biorxiv.org/content/10.1101/722397v3.full) of popular ECG peak detection algorithms and their performance on established benchmarks.

DominiqueMakowski:
@JohnDoenut That's great 😍 Looking forward to your PR in the next decade 😁

DominiqueMakowski:
@Tam-Pham It would be useful to get an overview of the existing pipelines for ECG cleaning (parameters of filters etc..). For instance, biosspy uses only simple filtering:

https://github.com/PIA-Group/BioSPPy/blob/e65da30f6379852ecb98f8e2e0c9b4b5175416c3/biosppy/signals/ecg.py#L69-L76

[...]

PR #65, opened 2020-01-03
Proposal for additional ECG functions

JohnDoenut:
# Description

Dear all,
Following #31, I would like to add some functions to the `ecg` module.

# Proposed Changes
Adhering to the structure of the `rsp` module, the following functions were added to the `ecg` module:

+ `ecg_clean()`
+ `ecg_findpeaks()`
+ `ecg_rate()`
+ `ecg_process()`
+ `ecg_plot()`

Note that for all these functios, examples are missing in the docstrings. Test for all these functions were added to `tests_ecg()`. Additionally, `ecg_findpeaks()` was recently benchmarked on the  [Glasgow University Database (GUDB)](http://researchdata.gla.ac.uk/716/) (see [5. Changelog](https://github.com/JohnDoenut/biopeaks) for details). Also note that `ecg_findpeaks()` is set up for the use of other peak detection algorithms that could be chosen using the `method` argument.

Please find a demo of  `ecg_process()` and `ecg_plot()` in /scripts/ecg_demo.ipynb.

Lastly, I made some minor modifications to the `signal` module in `signal_smooth()` and added a test for `signal_smooth()` to `tests_signal()`.


DominiqueMakowski:
😮💖 

give me some time to fully appreciate this 😁

JohnDoenut:
The failing test on `ecg_findpeaks()` is propably caused by the behavior of numpy.sum() which seems [non-uniform across architectures](https://stackoverflow.com/questions/41705764/numpy-sum-giving-strange-results-on-large-arrays).

On my Window 64bit machine the test doesn't fail, whereas on the Linux 64bit machine that runs the CI it does fail. It seems that numpy defaults to 32bit on Windows (64 and 32) whereas on Unix 64 systems it doesn't.

PR #66, opened 2020-01-04
Dev

DominiqueMakowski:


DominiqueMakowski:
There seems to be a lot of overlap between `rsp_rate` and `ecg_rate`, maybe we could put this in a `signal_rate` generic?

Issue #87, opened 2020-01-18
Docs: Tutorials/Example Ideas (and how to contribute)


DominiqueMakowski:
# Examples ideas

*Examples are short, specific and self-contained articles about specific topics or possibilities.*

- [x] **Understanding NeuroKit**: how to see what a function does in the docs, then its code on github, then where is the code located on your machine, and where you can try to make changes/fixes
- [x] **How to contribute**: once some change/fixes are mde, how to push back to github. With screenshots etc. (all steps from when github desktop is installed to PR)
- [x] **How to help with the documentation**: Pretty much what I wrote below, + also different types of documentation (i.e., examples, docstrings, etc) 
- [x] **Building a custom processing pipeline**: Where to see how a pipeline is made, and how to build a new function with a custom routine
- [x] **Get familiar with Python in 10min**: Crash course to python presenting the basics useful for using neurokit 
- [ ] **How to record ECG**: Documenting how to position electrodes etc.
- [ ] **Heart rate and artefacts detection**: About the peaks artefacts detection, visualization etc. (@JohnDoenut do you have some ideas?)
- [x] **Event-related analysis**: there is a basis in NK1's doc
- [x] **Interval-related analysis**: how to do get physiological features for longer periods (example with resting state)
- [x] **How to create epochs**: How to create epochs of data, either from one signal (showing events detection, and documenting the `epochs_create` function) or from multiple data (when people have different files for each subject), how to read them in a loop and put them in a `dict` and format it as epochs so it can be further analyzed.
- [x] **Extract and visualize heartbeats (QRS)**: Check if R peaks correctly detected, and adjust the method if need be. Then, extract and plot all R peaks and QRS complexes.
- [ ] **Find average respiration rate**: Get average heart rate and RSP rate (e.g., https://python-heart-rate-analysis-toolkit.readthedocs.io/en/latest/heartrateanalysis.html)
- [ ] **Heart Rate Variability (HRV)**: once we have HRV, write about cardiac coherence (what it is, what it isn't, how to estimate it) ([inspiration](https://www.kubios.com/about-hrv/))
- [x] **Electrodermal Activity (EDA) analysis**: phasic, tonic, SCR's features etc. (https://imotions.com/blog/skin-conductance-response/; https://support.mindwaretech.com/2018/04/all-about-eda-part-2-components-of-skin-conductance/)
- [ ] **Working with non-human data**: Showing another applied example of "building a custom pipeline" (#104)
[...]

PR #103, opened 2020-01-24
dev

DominiqueMakowski:


DominiqueMakowski:
@JohnDoenut should we rename `fixpeaks` to something else? `cleanpeaks` or something?

JohnDoenut:
> 
> 
> @JohnDoenut should we rename `fixpeaks` to something else? `cleanpeaks` or something?

@DominiqueMakowski , I had similar thoughts. Since the method technically doesn't apply to peaks, it could make more sense to call it `cleanibi` or `cleanperiods`. In that case it might also be worth considering to move the logic of the current `fixpeaks` implementation entirely to `ecg_rate`. The latter could then be called with an "autocorrect" flag or something.

DominiqueMakowski:
@JohnDoenut Mmh initially I was about to say that case we could have`*_findpeaks()` become `_peaks()` (only detection, but with a simpler name than *findpeaks*), to remove this not really relevant `fixpeaks()` step, but then it might still relevant for instance for EDA. So, in this case, yeah, I think it makes sense to move it somewhere to _rate as it is rather specific to the rate (ALTHOUGH detecting anomalies in the rate (the RRi) is built on the assumption that the artefact is in the peak location, hence wouldn't it make sense to fix/remove the artefacted peaks for instance for amplitude as well?).  

We could think of either `ecg_fixrate` or `ecg_cleanrate` that could be either an internal to `ecg_rate` or to expose it? Along with a `ecg_rate_plot()` that could plot the fixed and raw rates, as well as the quadrant diagram on the side (i.e., all plots related to rate and its fixing). What do you think about keeping `_fixpeak()` (mainly for API consistency in regards to other type of signals where it might be useful) and the eventual fixrate/cleanrate/rate_clean? 

(which makes me think that if we are to add functions of the form `something_something_plot()`, we might as well break current plots into subparts (e.g., `ecg_amplitude_plot()`, `ecg_clean_plot()`, ...), and then the main `ecg_plot()` would just gather all the subplots and combine them into one. This would allow users to get specific plots more easily if needed. Thoughts?)


On a side note, I enabled the support for jupyer notebooks for readthedocs, meaning that we will be able to include notebooks directly as examples, which will streamline a lot the example creation workflow (currently described in #87)   @Tam-Pham @zen-juen

JohnDoenut:
@DominiqueMakowski , After thinking a bit more about this I am back to wanting to keep `ecg_fixpeaks()` with the following change: the method would correct the R-peak indices directly and return the artifact indices as well as the corrected R-peak indices.
This means that we could remove the entire artifact correction logic from `ecg_rate()`. The reason I haven't done this before is that I assumed that the artifact correction could not be implemented such that it can be applied to the R-peak indices directly and that it would have to be applied to the rate instead. I think I know now how I could apply the artifact correction to the R-peak indices directly.
(removing extra indices, adding missed indices, and placing the remaining indices {ectopic, longshort} halfway beween their preceding and following neighbors).

If we can agree on this implementation I could refactor the affected methods at some point within the next two weeks.

Also, I don't think that consistency in method names (across modalities) needs to be enforced at all costs. There will necessarily be things that are unique to a specific modality and cannot be generalized.

Really nice with the .ipynb support for the documentation btw!

DominiqueMakowski:
GOAL: merge this monster-PR by tomorrow (Singapore time) 😁 

@sangfrois could you check the [docs](https://external-builds.readthedocs.io/html/neurokit2/103/tutorials/understanding.html) and see if it renders as you expect thanks!

sangfrois:
Of course. The docs look good. I'll make the proposed change.

PR #119, opened 2020-02-11
dev

DominiqueMakowski:


DominiqueMakowski:
Need to fix all instances where signal_interpolate() is used to adapt to new behaviour (https://github.com/neuropsychology/NeuroKit/pull/125)

DominiqueMakowski:
@zen-juen check my update on ecg_eventrelated, and check if it makes sense to also split the other functions similarly, i.e., one internal function per column.

Issue #122, opened 2020-02-13
rsp_amplitude returns wrong values? 


Mitchellb16:
I've been working to get an example/pipeline going for #104 , and ran into this issue with what seem to be the wrong values for amplitude. I looked into [rsp_amplitude](https://github.com/neuropsychology/NeuroKit/blob/master/neurokit2/rsp/rsp_amplitude.py) and I think that in spite of ensuring that each trough corresponds to its adjacent peak by making their lists equal length, they may have somehow become bumped in one direction or another? I'm not sure but I have 2 examples of it below. 

In both of the time segments it seems that there is a large amplitude breath near the end of the segment, but a larger amplitude is only seen in the preceding cycles, where they should not be.  

<img width="1052" alt="Screen Shot 2020-02-13 at 2 46 01 PM" src="https://user-images.githubusercontent.com/55215406/74472367-ce6fef00-4e6f-11ea-96a1-a01dbd86ee9d.png">

<img width="1050" alt="Screen Shot 2020-02-13 at 2 46 56 PM" src="https://user-images.githubusercontent.com/55215406/74472380-d760c080-4e6f-11ea-9491-167f79946877.png">

DominiqueMakowski:
Hi @Mitchellb16!

Indeed, seems like there is a problem, thanks for reporting. Could you try updating your version of neurokit (based on the plot I suspect that it is outdated, and I recall @JohnDoenut fixing the amplitude).

Try `pip uninstall neurokit2` and then `pip install https://github.com/neuropsychology/neurokit/zipball/master`

Let me know!

JohnDoenut:
@DominiqueMakowski , I believe that @Mitchellb16  is right. I noticed what @Mitchellb16 describes some time ago as well and always meant to look into it. After some initial poking around it seems that the bug is in `signal_interpolate()`. I think that we should simplify `signal_interpolate()` since I feel that the use of `x_axis` and `new_x` is somewhat convoluted.

Mitchellb16:
Hey @DominiqueMakowski, thanks for the help on this!  It seems the problem persists after updating. However, just to add some complexity, it seems when I crop the file to the 10 seconds surrounding this problem area, the values look more accurate, suggesting the problem likely occurred somewhere earlier in the file. 

Also now that I think about it, it looks like both rate and amplitude values have been "shifted" (compare with my second example above).

[...]

PR #125, opened 2020-02-14
Proposal to simplify signal interpolation

JohnDoenut:
Related to #122 I propose the commited changes to `signal_interpolate()`. I adapted the fucntions calling `signal_interpolate()` within the `rsp` module. The problem described in #122 seems fixed (confirmed by looking at plotted output of `rsp_process()` before and after the fix). Note however that other modules also rely on `signal_interpolate()` and that in those modules stuff probably breaks now.

### DIFF
--- a/neurokit2/signal/signal_interpolate.py
+++ b/neurokit2/signal/signal_interpolate.py
@@ -1,24 +1,28 @@
 # -*- coding: utf-8 -*-
 import numpy as np
-import scipy.interpolate
+from scipy.interpolate import interp1d
 
 
-def signal_interpolate(signal, desired_length=None, method="quadratic",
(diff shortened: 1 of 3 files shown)


neurokit2/signal/signal_interpolate.py
```diff
@@ -1,24 +1,28 @@
 # -*- coding: utf-8 -*-
 import numpy as np
-import scipy.interpolate
+from scipy.interpolate import interp1d
```
DominiqueMakowski: @JohnDoenut I prefer the first version as it is IMO always clearer to retain the module.submodule names throughout, even if it seems a bit more verbose. But it's easier to know from where it comes

PR #140, opened 2020-03-01
Bugfix: wrong ecg_fixpeaks keyword argument lists

hungpham2511:
# Description

Fix a small bug in the ecg_fixpeaks function, which cause CI to fail.


Tam-Pham:
to fix change of argument in ecg_fixpeaks

PR #143, opened 2020-03-03
feature EMG segmentation

GanshengT:
# Description

This PR aims at adding EMG segmentation

# Proposed Changes

Detailed explications are in the .ipynb scripts


Tam-Pham:
This is amazing! Thanks much @GanshengT . @DominiqueMakowski is away but we will look into integrating it into the package soonest 💃

PR #160, opened 2020-03-23
Dev

sangfrois:
So, if I pull upstream/dev, i'm removing my changes. I don't want that. Instead, I want upstream to pull my-fork/dev. But, doing that, there will merge conflicts on `epochs_create` even though they are the same. Plus, the binder add-on is removed along with my example.

Like you said, @DominiqueMakowski I could have PR to merge my changes directly in master, but I'm working on my github skills :)

Let's see where it takes us. Thanks for your feedback :)


sangfrois:
damn, everything is removed now.

sangfrois:
I'll try with another branch

PR #163, opened 2020-03-25
dev

DominiqueMakowski:


DominiqueMakowski:
I think we can also directly start adding the `*_analyze(data, sampling_rate, method="auto")` functions. Method can be either events/eventsrelated or interval/intervalrelated/period and if "auto" then selects the right method based on the mean duration of the data inputted (10s seems to be a sensible threshold)

DominiqueMakowski:
I've initialized, in the docs/img folder, a "workflow" graph that we can improve and display and the README and in the documentation. It aims at helping newcomers on which example/tutorial is adapted to their needs.

The design can definitely be improved (maybe using a different software than powerpoint), and for now we could, for instance, mark the non-available examples as "Work-in-progress". 

Clickable links would be great, but hard. One way of doing it would be to have a separate image for each link, and they'd be inserted side-by-side below the main flowchart, and these images would redirect to the appropriate URL.

workflow

DominiqueMakowski:
@Tam-Pham maybe it's possible to revise and streamline the QRS delineation example using the `show` arg? (also, need to rename the "derivative" method name in this example)

README.rst
```diff
@@ -98,11 +98,10 @@ Examples
 -  `Extract and Visualize Individual Heartbeats <https://neurokit2.readthedocs.io/en/latest/examples/heartbeats.html>`_
 -  `Delineate QRS peaks, onsets and offsets <https://neurokit2.readthedocs.io/en/latest/examples/ecg_delineation.html>`_
 
-*You can try out these examples directly in your browser by* [***clicking here***](https://github.com/neuropsychology/NeuroKit/tree/master/docs/examples#cloud-based-interactive-examples).
+*You can try out these examples directly in your browser by clicking on the badge here*
 
-
-.. image:: https://raw.github.com/neuropsychology/NeuroKit/master/docs/img/workflow.png
```
DominiqueMakowski: this one's not rendering properly because it hasn't yet been merged to master, but it's correct!

README.rst
```diff
@@ -98,11 +98,10 @@ Examples
 -  `Extract and Visualize Individual Heartbeats <https://neurokit2.readthedocs.io/en/latest/examples/heartbeats.html>`_
 -  `Delineate QRS peaks, onsets and offsets <https://neurokit2.readthedocs.io/en/latest/examples/ecg_delineation.html>`_
 
-*You can try out these examples directly in your browser by* [***clicking here***](https://github.com/neuropsychology/NeuroKit/tree/master/docs/examples#cloud-based-interactive-examples).
+*You can try out these examples directly in your browser by clicking on the badge here*
```
DominiqueMakowski: I think it's best to redirect them to the README here, reason is that if they know what's binder and all, they will know that they can click on the badge and will know what to expect. 

In the contrary case, they might benefit from first reading the documentation regarding the cloud-based interactive examples

PR #168, opened 2020-03-28
Dev

DominiqueMakowski:


DominiqueMakowski:
@Tam-Pham ok we're back with a working autodoc, there were some modules that we needed to mock (I don't remember why we removed them I think it was at the time when the docs were failing)

DominiqueMakowski:
Alright we can check briefly everything again, and then merge 💪

JanCBrammer:
#172 should fix the failing test

DominiqueMakowski:
interesting, it seems like it fails without failing on 3.6 @Tam-Pham any ideas?

DominiqueMakowski:
okay it got fixed when rerun, I guess travis works in mysterious ways 🤷‍♂

PR #192, opened 2020-04-09
Pre-compute means in _ecg_findpeaks_MWA

purpl3F0x:
# Description

Uses a (inclusive )scan to precompute the individual means of the signal. Instead of computing the a mean every time(Yields about x10 performance.)

# Proposed Changes

I changed the `_ecg_findpeaks_MWA `  in  `neurokit2/ecg/ecg_findpeaks.py`

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -30,6 +30,7 @@ Contributors
 * `@renatosc <https://github.com/renatosc>`_
 * `Nicolas Beaudoin-Gagnon <https://github.com/Fegalf>`_ *(Université Laval, Canada)*
 * `@rubinovitz <https://github.com/rubinovitz>`_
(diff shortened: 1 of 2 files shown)


DominiqueMakowski:
@purpl3F0x thanks a lot for this great performance improvement! Before I merge, you might want to add yourself to the [list of contributors](https://github.com/neuropsychology/NeuroKit/blob/master/AUTHORS.rst#contributors)?

(If by any chance you spot other places to do your magick (and there might be quite a lot 😁), do not hesitate to place yourself on the dev branch and make PRs to the dev - it's easier for us to manage then)

purpl3F0x:
Added 😎 

(You can always re-write it in c++ to improve the performance 🤣)

PR #195, opened 2020-04-17
dev (0.0.27)

DominiqueMakowski:


DominiqueMakowski:
@sangfrois I'll start revising. You have made nice (and lots) of additions and changes, which means my task is now to streamline and refocus if need be 😁 

It's always painful to see someone cut or rephrase something that we wrote (trust me I've been there :sweat_smile:), thus I'll try to make as many comments as I can but feel free to ask if you don't understand a particular change. 

I remember back in the days hating my mentor @marcosperduti when he used to cut "à la tronçonneuse" my papers, but in the end it was always for the best ☺️

sangfrois:
Hi @DominiqueMakowski 
Thanks for taking the time to reassure me that all of this is not personal. Maybe, my last interactions gave you the impression that I'd feel bad for getting revised .I was only worried changes I made would be forgotten about or erased unintentionally.

Please go ahead : cut, rephrase. Do all you think is needed to improve. As I did. I know it's in our best interest. I also have a mentor since my early days. He taught me that this process is all about caring for one another's commitment.

Please take your time, I know I wrote a lot and an ready to see it evolve, cut and rephrased

DominiqueMakowski:
about plotting epochs instead of arbiratry slice, it's true... could be a thing to consider. On the other hand, it would get completely crowded with all the new signals (as it comes after the processing), and would be actually quite complicated to read (4 panes with 15 lines each). And the plot is mostly here to make illustrate what a dataset of physiological signals look like, rather than showing something some important results or such... so IDK, we could give it a try and see how it looks like

DominiqueMakowski:
Merging this for now and will go back to the ms asap

sangfrois:
Hey Dominique! Thanks for reviewing and giving comments. 

About plotting epochs : I was rather suggesting we have one epochs or so with some measures and 1 or 2 cleaned signals. I get that plots are kind of accessory and that we're not showing important results. Though, I think it would be great to showcase `epochs_plot()` with selected measures. 

I thought that if we wanted to illustrate signals, we just might as well present them after processing and separating in epochs. IDK... I would've code the thing in the manuscript to illustrate what I mean, but I'm not savvy with R. I'm sorry about that.

DominiqueMakowski:
> Though, I think it would be great to showcase epochs_plot() with selected measures.

Yeay I agree, my only worry is that we'd have to make a selection in the signals to show (in the epochs) in order not to cram it. and this would add some additional code in between epoching and preprocessing (get the preprocessed data, select the columns to keep, epoch them and plot, but then re-get all of the original preprocessed data, epoch it, and then passed the full thing to the analysis)

> I would've code the thing in the manuscript to illustrate what I mean, but I'm not savvy with R.

Can try it in python to get us a better idea of what you mean! then for the plot I can maybe reproduce in R if need be (it's not necessary tho, it's just that I personally *hate* matplotlib) :)

sangfrois:
Does that make any more sense to you ?
```
# Load the package
import neurokit2 as nk

# Download example dataset
data = nk.data("bio_eventrelated_100hz")

# Process the data
df, info = nk.bio_process(ecg=data["ECG"],
                          rsp=data["RSP"],
                          eda=data["EDA"],
                          sampling_rate=100)

# Find events
conditions = ["Negative", "Neutral", "Neutral", "Negative"]
events = nk.events_find(event_channel=data["Photosensor"],
                        threshold_keep='below',
                        event_conditions=conditions)

# Epoch the data
epochs = nk.epochs_create(data=df,
                          events=events,
                          sampling_rate=100,
                          epochs_start=-0.1,
                          epochs_end=4)
# select subset
subset = epochs[some_idx or range]

# HERE, show subset
fig = nk.epochs_plot(subset)

# Extract event related features
results = nk.bio_analyze(epochs)

# Show subset of results
results[["Condition", "ECG_Rate_Mean", "RSP_Rate_Mean", "EDA_Peak_Amplitude"]]

DominiqueMakowski:
you mean a selection of epochs or of the signals in the epochs?

sangfrois:
Both, in fact. Sorry I sent comment too rapidly

PR #196, opened 2020-04-17
Changes to Welch method in signal_psd()

Tam-Pham:
# Proposed Changes

To determine the window_length, the following steps were made:

## Prioritize frequency resolution
A nperseg long enough to capture at least 2 cycles of lowest frequency:
```
# Define window length
        if min_frequency == 0:
            min_frequency = 0.001  # sanitize lowest frequency
        if window is not None:
            nperseg = int(window * sampling_rate)
        else:
            # to capture at least 2  cycles of min_frequency
            nperseg = int((2 / min_frequency) * sampling_rate)
```

However, since our signals are mostly sampled at high frequency (1000-4000Hz) and min_frequency is often very low (~0 Hz 😅 ), nperseg can be super huge, sometimes longer than duration of signal. 
 
## Sacrifice frequency resolution for more appropriate window length

```
# in case duration of recording is not sufficient
        if nperseg > len(signal) / 2:
            print("Neurokit warning: signal_psd(): The duration of recording is too short to support a sufficiently long window for high frequency resolution. Consider using a longer recording or increasing the `min_frequency`")
            nperseg = int(len(signal / 2))
            nperseg = nperseg * resolution
```
Here, an argument resolution is used to further adjust the nperseg. In cases where nperseg is still too large and temporal resolution is jeopardized (too many variances captured, often seen in very sharpy, pointy power plot), users can choose to cut nperseg further by decreasing the resolution (frequency resolution). 

For example, using an ecg_signal of 4 minutes, sampling_rate=4000Hz and min_frequency=0Hz, at default resolution of 0.5:
image

and cut the resolution to 0.3: 
image

The balance is very much depending on the user's decision.
P/S: the Power axis is now not logged, current unit is (ms^2/Hz), which is a rather common unit. Other scales that are common are (s^2/Hz) or decibels (dB = 10 * log10(power)). We can pick one or let the user decides 😃

### DIFF
--- a/neurokit2/ecg/ecg_hrv.py
+++ b/neurokit2/ecg/ecg_hrv.py
@@ -155,7 +155,7 @@ def _ecg_hrv_time(rri):
 
 
 def _ecg_hrv_frequency(ecg_period, sampling_rate=1000, ulf=(0, 0.0033), vlf=(0.0033, 0.04), lf=(0.04, 0.15), hf=(0.15, 0.4), vhf=(0.4, 0.5), method="welch", show=False):
-    power = signal_power(ecg_period, frequency_band=[ulf, vlf, lf, hf, vhf], sampling_rate=sampling_rate, method=method, max_frequency=0.5, resolution=0.3, show=show)
+    power = signal_power(ecg_period, frequency_band=[ulf, vlf, lf, hf, vhf], sampling_rate=sampling_rate, method=method, max_frequency=0.5, show=show)
     power.columns = ["ULF", "VLF", "LF", "HF", "VHF"]
     out = power.to_dict(orient="index")[0]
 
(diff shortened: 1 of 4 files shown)


DominiqueMakowski:
> The balance is very much depending on the user's decision.

But do we have the best possible default?

Tam-Pham:
@DominiqueMakowski I would say 1 aka a redundant parameter. 

I don't think it's necessary to further adjust the nperseg by resolution as the previous calculations are already following best practice out there. But I added it anyway as a mean to customize nperseg if need be. 

But it's definitely something that can be removed

DominiqueMakowski:
Well I think we can hide this argument as it is not something users should tweak too much as it could just adds confusion and variability for analyses where it used. So we can just add a [args](https://www.geeksforgeeks.org/args-kwargs-python/) argument that is propagated into the Welch function

Tam-Pham:
ok can! I will do that. 

How about the unit of the spectrum?

Without log (like the one in the above figs), the scale is  (ms^2/Hz), one of the common ones. Other scales that are common are (s^2/Hz) or decibels (dB = 10 * log10(power)). We can pick one or let the user decides (e.g: db=False/True). 

What chu think? @DominiqueMakowski

DominiqueMakowski:
I think we set it to the most common then, without log

PR #197, opened 2020-04-22
dev

DominiqueMakowski:


DominiqueMakowski:
@sangfrois I removed the "step1, step2" in the analysis as it was short enough to skip it, as well as the HRV chunk as it is included in the analysis. I also cut or rephrased some of the more theoretical elements, as it was sometimes a bit off-topic as for that paper ☺️ Last point to eventually modify would be the figure, but we'll see, as it would require some more code to make it look nice and digest, but that would complexify the code. Let's keep this in mind tho in case we have some ideas.

sangfrois:
I'll have a look at it! Thanks for keeping me in the loop. I understand that some ideas were off topic or the tone, a bit harsh. Thanks for reviewing, and commenting, it was always makes us better :)

PR #201, opened 2020-05-06
dev

DominiqueMakowski:


tests/tests_ecg.py
```diff
@@ -3,7 +3,7 @@
 import neurokit2 as nk
 import matplotlib.pyplot as plt
 
-import biosppy
+# import biosppy
```
DominiqueMakowski, line 6: I think this needs to be there so that it's tested against, + these packages are only downloaded for the tests on travis & co, and are not part of dependencies

DominiqueMakowski:
There's something fishy, iit seems that the CI is lagging behind, I fixed the tests two commits ago and yet it still fails 🤔

DominiqueMakowski:
Okay that's it Im merging despite of travis we'll see how

PR #212, opened 2020-05-17
0.0.32

DominiqueMakowski:

### DIFF
--- a/neurokit2/signal/signal_fixpeaks.py
+++ b/neurokit2/signal/signal_fixpeaks.py
@@ -371,14 +371,10 @@ def _correct_misaligned(misaligned_idcs, peaks):
     # the total number of peaks. prev_peaks and next_peaks must have the
     # same number of elements.
     valid_idcs = np.logical_and(misaligned_idcs > 1,
-                                misaligned_idcs < len(corrected_peaks))
+                                misaligned_idcs < len(corrected_peaks) - 1)
     misaligned_idcs = misaligned_idcs[valid_idcs]
-    prev_peaks = corrected_peaks[[i - 1 for i in misaligned_idcs if i - 1 > 0]]
-    next_peaks = corrected_peaks[[i + 1 for i in misaligned_idcs if i + 1 < len(corrected_peaks)]]  # make sure i + 1 does not exceed total peaks number
-
-    if len(next_peaks) < len(prev_peaks):
-        # Drop last index of prev_peaks if unequal number of prev and next peaks
-        prev_peaks = prev_peaks[0:len(next_peaks)]
+    prev_peaks = corrected_peaks[[i - 1 for i in misaligned_idcs]]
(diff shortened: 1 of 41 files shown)


neurokit2/signal/signal_fixpeaks.py
```diff
@@ -373,8 +373,13 @@ def _correct_misaligned(misaligned_idcs, peaks):
     valid_idcs = np.logical_and(misaligned_idcs > 1,
                                 misaligned_idcs < len(corrected_peaks))
     misaligned_idcs = misaligned_idcs[valid_idcs]
-    prev_peaks = corrected_peaks[[i - 1 for i in misaligned_idcs]]
-    next_peaks = corrected_peaks[[i + 1 for i in misaligned_idcs]]
+    prev_peaks = corrected_peaks[[i - 1 for i in misaligned_idcs if i - 1 > 0]]
```
JanCBrammer, line 376: @zen-juen, note that the conditions that you are evaluating in the list comprehensions are already evaluated in `valid_idcs`.

zen-juen: @JanCBrammer I made the changes because using actual data next_peaks was not computable because i+1 exceeded the index limit for misaligned_idcs

let me check again!

JanCBrammer: You're right. Simply change `len(corrected_peaks)` to `len(corrected_peaks) - 1` in `valid_idcs`. Then you can remove the rest of the changes. Nice catch!

DominiqueMakowski:
> but also convention

Is it a convention to sample at 10 Hz for HRV?

> It would be good to enforce a minimal sampling rate though (in cases where the original sampling rate is too low to estimate the vhf component).

We could say by default we resample at the original sampling rate, but if it's below < 10 we resample at 10. We just need to pass this information to `signal_psd()`

JanCBrammer:
> Is it a convention to sample at 10 Hz for HRV?

Yes, most implementations use 4, 7, or 10Hz.


> We could say by default we resample at the original sampling rate, but if it's below < 10 we resample at 10. We just need to pass this information to `signal_psd()`

+1

neurokit2/signal/signal_fixpeaks.py
JanCBrammer: I also realize that we can simplify https://github.com/neuropsychology/NeuroKit/blob/3f5efda8cd98e35c71aaa44cd80afb6dda87182d/neurokit2/signal/signal_fixpeaks.py#L354-L355

to `valid_idcs = missed_idcs > 1` since the indices are not incremented in `_correct_missed()`

DominiqueMakowski:
@Tam-Pham I guess travis is back 😬

DominiqueMakowski:
before merging:
- [ ] `complexity_plot()` (we can make a separate function for it, and additionally return it if `show=True` in `complexity_optimize()`
- [ ] `hrv_plot()` same, we can expose it and return it if need be in `hrv()`

PR #215, opened 2020-05-19
Bug on fixpeaks

TiagoTostas:
I think there is a bug on a sign according to the original paper, could you check please?


neurokit2/signal/signal_fixpeaks.py
```diff
@@ -225,8 +225,8 @@ def _find_artifacts(peaks, c1=0.13, c2=0.17, alpha=5.2, window_width=91,
         if np.abs(drrs[i]) <= 1:    # Figure 1
             i += 1
             continue
-        eq1 = np.logical_and(drrs[i] > 1, s12[i] < (-c1 * drrs[i] - c2))    # Figure 2a
-        eq2 = np.logical_and(drrs[i] < -1, s12[i] > (-c1 * drrs[i] + c2))    # Figure 2a
+        eq1 = np.logical_and(drrs[i] > 1, s12[i] < (-c1 * drrs[i] + c2))    # Figure 2a
```
DominiqueMakowski, line 228: pinging @JanCBrammer

DominiqueMakowski:
@TiagoTostas you might want to re-open this PR so that it leads to the dev branch instead of the master if we don't wanna create a GitHub Ragnarok ☺️

neurokit2/signal/signal_fixpeaks.py
TiagoTostas: Thank you @JanCBrammer! 
Those equations make sense with Fig2a indeed ( I was careless and only following the decision tree) 👍

PR #216, opened 2020-05-20
fix variable-patches

TiagoTostas:
As discussed with Jan,
- Fixed patches of the polygon described in the paper (better visualization).  [picture]
- Adjust limits to the threshold plots.

polygonpatches


JanCBrammer:
Nice, thanks @TiagoTostas!

PR #232, opened 2020-06-01
0.0.36

DominiqueMakowski:
Focus on improving testing and general worfklow and stability.


neurokit2/stats/mutual_information.py
```diff
@@ -164,8 +164,8 @@ def _entropy(X, k=1):
     return d*mean(log(r))+log(volume_unit_ball)+log(n-1)-log(k)
     """
     return (
-        d * np.mean(np.log(r + np.finfo(X.dtype).eps))
-        + np.log(volume_unit_ball)
-        + scipy.special.psi(n)
-        - scipy.special.psi(k)
+        d * np.mean(np.log(r + np.finfo(X.dtype).eps)) +
```
DominiqueMakowski, line 167: it's possible that black changes this, if it does, let's remove this W503 check from flake8

DominiqueMakowski:
Because we changed the docs structure, we need to check and update all the links (including the README), the tricky part is that because the new website is not built, the new links don't work yet, so we have to "guess" them

DominiqueMakowski:
I think we're ready to merge, as the massive main overhaul in terms of cleaning, fixing and setup-ing all the linting checks and all have been done 😩 that was tiring but needed to be done ☺️

PR #233, opened 2020-06-02
Exploring Github CI

Tam-Pham:
- [x] Add extras_require in setup.py for test deps 

- [x] Use extras_require to import test deps in tox.ini

- [x] move doctest to tox.ini

- [x] remove pytest from Github CI Tests (only test with tox)

- [x] add doctest to Github CI


.github/workflows/tests.yml
```diff
@@ -22,7 +22,7 @@ jobs:
         #    python-version: 3.8
 
     steps:
-    - uses: actions/checkout@master
+    - uses: actions/checkout@v2
```
DominiqueMakowski, line 25: my guess is that master is always up-to-date whereas V2 might get out-dated when V3 comes out

Tam-Pham: yeahh, i was just trying to see if it makes any difference. Reverting back to master

hungpham2511: master is the latest development and might contain bugs. So normally its preferable to use a released and numbered version like v2.

DominiqueMakowski: ok, but that means from time to time we'll need to check if there is a V3 out? we can change to V2 for checkout and the other action in all of the workflows

PR #258, opened 2020-06-15
0.0.38

DominiqueMakowski:
## Goals

- Try a `PR-per-feature to dev` workflow


DominiqueMakowski:
FYI @sangfrois @JanCBrammer we started trying out a new workflow, basically rather than committing directly to dev we create new PRs (directed to dev), and we merge them once its green (so that dev is also quite stable). 

This way, although tedious cause we need to manage more branches and PR, but at least every PR has a narrower scope. 

Also, PR-based changes are nicely and automatically gathered by GitHub every time dev is merged to master (see https://github.com/neuropsychology/NeuroKit/releases) so that's cool.

JanCBrammer:
> rather than committing directly to dev we create new PRs (directed to dev), and we merge them once its green (so that dev is also quite stable).

@DominiqueMakowski , noted! So basically PRs from feature branches in our own forks?

DominiqueMakowski:
@JanCBrammer not necessarily, what I do is (on github desktop) I set myself on the dev branch, from there create a new branch, select 'based on dev', and then publish this branch and make a pull request back to dev (e.g., https://github.com/neuropsychology/NeuroKit/pull/269) so my guess is that it doesn't need to be from your own fork since you have direct writing rights

docs/examples/custom.ipynb
```diff
@@ -25,6 +25,7 @@
     "import numpy as np\n",
     "import matplotlib.pyplot as plt\n",
     "import pandas as pd\n",
+    "%matplotlib inline\n",
```
Tam-Pham, line 28: hi @sangfrois, apparently, according to a discussion here https://github.com/spatialaudio/nbsphinx/issues/447#issuecomment-630766064, the line `%matplotlib inline` isn't needed anymore in modern IPython versions. Did it not work for you?

sangfrois: Hi @Tam-Pham, so the problem is that you get a Javascript error in the Binder environment. ``%matplotlib inline`` fixes that problem in the cloud examples. So, yes I can run the notebooks without this command on my local machine, but not using the link in the README under ``docs/examples".

Does that make any sense ? :)

sangfrois:
Hey @DominiqueMakowski, right! I saw that just after pushing the minor fix for the python notebooks. Maybe this could be added in the "how to contribute" section of our docs.

>  so my guess is that it doesn't need to be from your own fork since you have direct writing rights

From what I get : I could open a new branch based on ``upstream/dev``. It does not have to be my fork (i.e. ``origin/dev``). 

 This new branch contains my new feature (``upstream/sangfrois-feat``), then when all checks pass I can put it against ``upstream/dev``. 
 
 After reviewing changes, one merge from ``upstream/sangfrois-feat`` to ``upstream/dev``, then a second merge from ``upstream/dev`` to ``upstream/master``

DominiqueMakowski:
Let's merge this bad boi

PR #269, opened 2020-06-16
[feature] Blink detection improvements

DominiqueMakowski:


DominiqueMakowski:
@zen-juen I'd say we merge this and then we can work on `_analyze()` and events delineation etc. later?

Issue #267, opened 2020-06-16
More PSD methods


DominiqueMakowski:
### Fourier-based methods
- [x] Welch (wrapper around Scipy)
- [x] multitapers (wrapper around mne)
- [x] Lomb-Scargle (in [pyHRV](https://github.com/PGomes92/pyhrv#frequency-domain-parameters) - around scipy)
### Minimum of variance MV (Capon)
- [ ] Capon
### Autoregressive (AR) parameters
- [x] burg (in [spectrum](https://github.com/cokelaer/spectrum))
- [ ] Yule-Walker  (in [pyHRV](https://github.com/PGomes92/pyhrv#frequency-domain-parameters) and [spectrum](https://github.com/cokelaer/spectrum))
- [ ] least squares linear prediction

*note: spectrum seems to have problems of installation*

PR #273, opened 2020-06-17
[feature] add new method for signal_psd()

Tam-Pham:
#267


neurokit2/signal/signal_psd.py
```diff
@@ -125,3 +109,199 @@ def signal_psd(
...
+            normalization="full",
+            verbose=False,
+        )
+    except ImportError:
+        raise ImportError(
+            "NeuroKit warning: signal_psd(): the 'mne'",
+            "module is required for the 'mne' method to run.",
+            "Please install it first (`pip install mne`).",
+        )
+    return power, frequency
```
DominiqueMakowski, line 137: note that in all of the methods you return frequency, power and here the other way around. not sure if intentional ^^

Tam-Pham: yeahh, i changed this to standardize with others =))

Issue #247, opened 2020-06-09
Improving the signal_interpolate() API.


Tam-Pham:
While the issue of signal_interpolation() is discussed in #242 and #243, I also noticed that in `_signal_formatpeaks_sanitize()`, the `desired length` for interpolation is set as `len(peaks_signal)`, which would be the length of the entire signal. 
https://github.com/neuropsychology/NeuroKit/blob/a8d2981155b031e88188b126e3120c74730bf792/neurokit2/signal/signal_formatpeaks.py#L71

However, in its use case in `signal_period()`, where `period` is interpolated between `peaks`, shouldn't the `desired_length` be the length between the `first_peak` and `last_peak` instead of the length of the entire signal?

@JanCBrammer @DominiqueMakowski. What do you guys think?

JanCBrammer:
@Tam-Pham, that is a good point and it made me review `signal_interpolate()` again. It would be ok to set `desired length` to the length of the signal if `signal_interpolate()` could extrapolate y values before the first peak (x_val[0]) and after the last peak (x_val[-1]). However, there are two issues with `signal_interpolate()` as it is currently implemeted:

1. extrapolation before x_val[0] or after x_val[-1] is not possible. This used to be possible with the `new_x` argument which I removed some time ago (my bad).

2. the scipy implementation of the Akima1DInterpolator does not allow for extrapolation. We need to use [monotonic cubic spline](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.PchipInterpolator.html) instead. The latter ensures monotone interpolation (see #243) and also allows extrapolation beyond x_val[0] and x_val[-1]. Again, my bad, I didn't realize that `akima` cannot extrapolate.

I won't have time today to fix this, but we need to adress this asap since a lot of crucial functionality depends on `signal_interpolate()`.

DominiqueMakowski:
Agreed with @JanCBrammer, good catch @Tam-Pham !

JanCBrammer:
I just pushed a [fix of `signal_interpolate()`](https://github.com/neuropsychology/NeuroKit/pull/244/commits/394262f7ee89d45b291db4e318ebe7030dfaa90e). Please review. Now it's a matter of making sure that an appropriate `new_x` is passed in, which basically is everywhere we call `signal_period()` or `signal_rate()`.

JanCBrammer:
[...]

PR #276, opened 2020-06-20
[API] API of signal_period() and signal_rate()

Tam-Pham:
## To continue discuss and finalize our discussion in #247 

We last agreed to **remove** `peaks` argument and **add** `signal` to `signal_period()` and `signal_rate()`.  And within these functions, `peaks` will be identified with `signal_findpeaks()`. 

However, on 2nd thought, I think `signal_findpeaks()` is a very general function. We have more specialized functions, for instance `ecg_peaks()` for different signal. For instance, here:

https://github.com/neuropsychology/NeuroKit/blob/65147605838947aa609e8116bfd5e80a60d5f419/neurokit2/ecg/ecg_process.py#L98

And hence, completely removing the `peaks` argument might not be the best solution imo (unless we want to specify the type of find_peaks() functions in `signal_period()` 🤔).

I made the changes in this PR to add signal (without removing peak) in `signal_period()` and `signal_rate()`. However, a problem I encountered is with `rsp_rrv()`

https://github.com/neuropsychology/NeuroKit/blob/65147605838947aa609e8116bfd5e80a60d5f419/neurokit2/rsp/rsp_rrv.py#L219

I don't know what to pass into this function as `signal` 😅 

What do u guys think? @JanCBrammer @DominiqueMakowski


JanCBrammer:
> I think `signal_findpeaks()` is a very general function. We have more specialized functions, for instance `ecg_peaks()` for different signal.

Looks like I haven't thought this through :grin:

> completely removing the `peaks` argument might not be the best solution

@Tam-Pham, agreed! Maybe it's best to stick to our original idea: Only have the `peaks` argument and by default not interpolate between `peaks`. Interpolation must be requested by setting `desired_length`. After all, it's only about making sure that nothing breaks downstream when `desired_length` is left at the default `None`.

It feels like having both arguments, `signal` and `peaks` could be confusing and also makes the implementation unnceccesarily convoluted (since, as you already pointed out, it would require passing the signal modality to a function that is supposed to be modality-agnostic).

Tam-Pham:
@JanCBrammer yes, I agree that having both `signal` and `peaks` argument is too clunky (signal is there just for its length!) 
> Maybe it's best to stick to our original idea: Only have the peaks argument and by default not interpolate between peaks. Interpolation must be requested by setting desired_length. After all, it's only about making sure that nothing breaks downstream when desired_length is left at the default None.

Okay! Let's stick to our original idea 😁

JanCBrammer:
@Tam-Pham, if you have no further additions I think we can merge.

Tam-Pham:
Superb! @JanCBrammer 
Let's merge this 😁

PR #289, opened 2020-06-28
0.0.39

DominiqueMakowski:


JanCBrammer:
@DominiqueMakowski, only one minor content issue with the current draft (version 3): NeuroKit cannot be used for biofeedback (as stated in the discussion), since it does not provide real-time utilities. I removed "biofeedback" in my latest edit.

DominiqueMakowski:
oops, we'll address that in between the reviews or before next submission (as it has been submitted)

I wonder what would be needed for our algorithms to make them run online though

JanCBrammer:
Real-time signal processing requires the data to be analyzed in small(ish) chunks while they are streaming from the acquisition device. That requires a dedicated API for communicating with devices, setting up event loops etc. Most importantly it would require us to re-write most of our algorithms since those are geared towards offline analyses (i.e., processing the entire signal at once). IMHO, real time processing is definitely out of scope for NeuroKit and there is a dedicated Python library for this already (https://github.com/eegsynth/eegsynth).

Nice that the paper is submitted btw!

DominiqueMakowski:
I *think* we can merge

PR #299, opened 2020-06-30
[feature] Add EOG into bio_* functions

zen-juen:
- [x] EOG into bio_process
- [x] EOG into bio_analyze
- [ ] Finalize EOG features to be returned in event-related / interval-related analysis ?


zen-juen:
for now, `eog_eventrelated()` returns:

- EOG_N_Blinks: Number of blink peaks during each epoch.
- EOG_Rate_Baseline: the baseline EOG rate before stimulus onset.
- EOG_Rate_Max: the maximum EOG rate after stimulus onset.
- EOG_Rate_Min: the minimum EOG rate after stimulus onset.
- EOG_Rate_Mean: the mean EOG rate after stimulus onset.
- EOG_Rate_Max_Time: the time at which maximum EOG rate occurs.
- EOG_Rate_Min_Time: the time at which minimum EOG rate occurs.


and  `eog_intervalrelated()` returns only:
- EOG_Rate_Mean

DominiqueMakowski:
I'd move N_blinks to intervalrelated() (and make its name consistent with SCR_ number of peaks), and keep only a binary var in eventrelated() whether there is at least blink in the epoch (between [0::], not in the baseline period) or not.

zen-juen:
> I'd move N_blinks to intervalrelated() (and make its name consistent with SCR_ number of peaks), and keep only a binary var in eventrelated() whether there is at least blink in the epoch (between [0::], not in the baseline period) or not.

Renamed to `EOG_Peaks_N` and `EOG_Blinks_Presence` in `eog_intervalrelated()` and `eog_eventrelated()` respectively

zen-juen:
@DominiqueMakowski  mergeeee?

PR #300, opened 2020-07-01
[Fix] PSD Methods

Tam-Pham:
- [x] fix hrv_plot function in hrv_frequency: method is not passed
- [x] check burg implementation
- [x] `hrv_frequency`: pass psd_method to plot correctly 
- [x] fix lomb implementation


zen-juen:
In `nk.signal_psd` function of the welch implementation

```
def _signal_psd_welch(
    signal, sampling_rate=1000, nperseg=None
):

    frequency, power = scipy.signal.welch(
        signal,
        fs=sampling_rate,
        scaling="density",
        detrend=False,
        nfft=int(nperseg * 2),
        average="mean",
        nperseg=nperseg,
    )
    return frequency, power

```

There is no argument to allow for overlapping data segments, which was originally specified as in `noverlap` (int, optional) in the scipy.signal.welch
```
scipy.signal.welch(x, fs=1.0, window='hann', nperseg=None, noverlap=None, nfft=None, detrend='constant', return_onesided=True, scaling='density', axis=- 1, average='mean')
```
I realized this while trying to implement the EDAsymp (which is in essence a PSD computation of the EDA signal and computes based on overlapping segments) 😄 Shall I add this `noverlap` argument in our welch implementation?

DominiqueMakowski:
Can make it pass through kwargs?

zen-juen:
initiated `eda_sympathetic()` in this pull request due to `signal_psd` dependency

DominiqueMakowski:
We do:

https://github.com/neuropsychology/NeuroKit/blob/1bb55cca9c10818e558e39afde42d59907ea3be6/neurokit2/hrv/hrv_utils.py#L22

But the `/ sampling _rate * sampling_rate` seems useless isn't it?

DominiqueMakowski:
Also, I'd say that we should fix the output of `signal_psd()` as follows:
```python
    if show is True:
        ax = data.plot(x="Frequency", y="Power", logy=True, title="Power Spectral Density (ms^2/Hz)")
        ax.set(xlabel="Frequency (Hz)", ylabel="Spectrum")
    return data
```
So that it always return the data rather than the plot regardless of the show parameters. Also, we should make `show` default to False (but explicitly turn it to true in the docstring examples)

Tam-Pham:
> We do:
> 
> https://github.com/neuropsychology/NeuroKit/blob/1bb55cca9c10818e558e39afde42d59907ea3be6/neurokit2/hrv/hrv_utils.py#L22
> 
> But the `/ sampling _rate * sampling_rate` seems useless isn't it?
> 
> @Tam-Pham

Hmm, yeahh it will be cancelled out 😅 Do you remembered who implemented it?

neurokit2/hrv/hrv_frequency.py
```diff
@@ -53,8 +53,8 @@ def hrv_frequency(
         If False, warnings will be printed. Default to True.
     show : bool
         If True, will plot the power in the different frequency bands.
-    norm : bool
-        Normalization of power.
+    normalization : bool
+        Normalization of power. Default to True.
```
DominiqueMakowski, line 57: @Tam-Pham could you add a tinybitsy more details here like why would someone want to normalize or not

neurokit2/signal/signal_power.py
```diff
@@ -22,8 +22,8 @@ def signal_power(signal, frequency_band, sampling_rate=1000, continuous=False, s
         Compute instant frequency, or continuous power.
     show : bool
         If True, will return a Poincaré plot. Defaults to False.
-    norm : bool
-        Normalization of power.
+    normalization : bool
```
DominiqueMakowski, line 25: same

neurokit2/signal/signal_psd.py
```diff
@@ -5,7 +5,19 @@
...
+    signal,
+    sampling_rate=1000,
+    method="welch",
+    show=False,
+    normalization=True,
+    min_frequency=0,
+    max_frequency=np.inf,
+    window=None,
+    window_type='hann',
+    ar_order=16,
```
DominiqueMakowski, line 17: I'd rename `ar_order` by `order` (so it's consistent with the following args) and mention that it's used for autoregressive methods

DominiqueMakowski:
> Hmm, yeahh it will be cancelled out 😅 Do you remembered who implemented it?

It got changed several times by myself and @JanCBrammer, so it's likely to be an omission. let's fix it

JanCBrammer:
> Do you remembered who implemented it?

Well, that's what `View git blame` is for I guess :D

neurokit2/signal/signal_psd.py
```diff
@@ -194,21 +221,38 @@ def _signal_psd_welch(signal, sampling_rate=1000, nperseg=None):
...
+#    # Specify sample times
+#    t = np.arange(len(signal))
+#
+#    power = np.asarray(scipy.signal.lombscargle(t, signal, frequency, normalize=True))
+    try:
+        from astropy.timeseries import LombScargle
+        if max_frequency == np.inf:
+            max_frequency = sampling_rate / 2  # sanitize highest frequency
+        t = np.arange(len(signal)) / sampling_rate
+        frequency, power = LombScargle(t, signal, normalization='psd').autopower(minimum_frequency=min_frequency, maximum_frequency=max_frequency)
```
DominiqueMakowski, line 246:
```suggestion
        import astropy.timeseries
        if max_frequency == np.inf:
            max_frequency = sampling_rate / 2  # sanitize highest frequency
        t = np.arange(len(signal)) / sampling_rate
        frequency, power = astropy.timeseries.LombScargle(t, signal, normalization='psd').autopower(minimum_frequency=min_frequency, maximum_frequency=max_frequency)
```

I don't like `from x import y` 😬

neurokit2/signal/signal_psd.py
```diff
@@ -8,14 +8,16 @@ def signal_psd(
     signal,
     sampling_rate=1000,
     method="welch",
-    show=True,
+    show=False,
+    normalization=True,
```
DominiqueMakowski, line 12: I'd call it `normalize` rather than `normalization` so that's consistent with the standardize function etc.

Tam-Pham:
> Well, that's what `View git blame` is for I guess :D

hahaha yeahhh but not to blame but to check with the implementer 😉

PR #306, opened 2020-07-05
[improvement] ECG Benchmarking: add 2 databases

DominiqueMakowski:


studies/hrv_frequency/make_data.py
```diff
@@ -0,0 +1,45 @@
...
+            sampling_rate = np.unique(data_participant["Sampling_Rate"])[0]
+            rpeaks = data_participant["Rpeaks"].values
+
+            # Interpolate
+            rri = np.diff(rpeaks) / sampling_rate * 1000
+            desired_length = int(np.rint(rpeaks[-1] / sampling_rate * sampling_rate))
+            rri = nk.signal_interpolate(rpeaks[1:], rri, x_new=np.arange(desired_length))
+
+            # Get PSD
+            psd = nk.signal_psd(rri, sampling_rate=sampling_rate)
```
DominiqueMakowski, line 32: @Tam-Pham I've initialized this script. Once PSD and HRV_freq is fixed, we can run this to basically extract the PSDs for all subjects of the databases using different methods, and compute the average pattern AND variability for each method so we can see how it looks like (and also investigate the optimal boundaries between HF LF etc.).

Issue #309, opened 2020-07-06
ECG delineate CWT mode crashes if no consecutive T and P peaks are found


raimonpv:
**Describe the bug**
```
Traceback (most recent call last):
  File "*/neurokit2/ecg/ecg_delineate.py", line 121, in ecg_delineate
    waves = _ecg_delineator_cwt(ecg_cleaned, rpeaks=rpeaks, sampling_rate=sampling_rate)
  File "*/neurokit2/ecg/ecg_delineate.py", line 468, in _ecg_delineator_cwt
    tpeaks, ppeaks = _peaks_delineator(ecg, rpeaks, sampling_rate=sampling_rate)
  File "*/neurokit2/ecg/ecg_delineate.py", line 625, in _peaks_delineator
    tpeaks, ppeaks = zip(*[(g[0], g[-1]) for g in significant_peaks_groups])
  File "*/neurokit2/ecg/ecg_delineate.py", line 625, in <listcomp>
    tpeaks, ppeaks = zip(*[(g[0], g[-1]) for g in significant_peaks_groups])
IndexError: list index out of range
```

**To Reproduce**
1. Run ecg_delineate function with "cwt" mode with an ECG signal with no consecutive T and P peaks.

**Expected behaviour**
Right now it is considered that the first peak found in _find_tppeaks is a T-peak, the last one a P-peak. If just one peak is found, it is considered that the T and P peaks are the same and if no peak is found it is returning an empty list that makes crash the `_peaks_delineator` function. I expect to return a Nan instead in order that the function doesn't crash and the resulting arrays of P and T peaks is the same length as the R peaks.
 
**Specifications**
- System: Linux
- Python version: 3.7.7
- NeuroKit Version: 0.0.37

Tam-Pham:
Thanks @raimonpv for reporting the bug. We will fix it asap and follow up here. 

In the meantime, if needed for your analysis, please rely on the `dwt` or `peaks` methods in `ecg_delineator()`.

raimonpv:
@Tam-Pham I had already fixed it and can make a PR with the code tomorrow morning if it's okey with you

DominiqueMakowski:
@raimonpv by all means, please PR :)

Tam-Pham:
That’s fantastic! We can further discuss the fixes in your PR. 💪🏻💪🏻💪🏻 looking forward to it

raimonpv:
Closed by #314

PR #314, opened 2020-07-07
[Fix] ecg_delineate CWT mode crash when no consecutive tp peaks found

raimonpv:
# Description
Fix ecg_delineate bug described in neuropsychology/NeuroKit#309

# Proposed Changes

The bug reported is fixed in line 668-669 by returning a list with a np.nan instead of an empty list in `_find_tppeaks`. Additionally, I changed the type of array in `_peaks_delineator` and in `_onsets_offsets_delineator` to object as a an array with np nans cannot be converted to int. Finally, in `_onsets_offsets_delineator` if the peak is a np.nan now it appends a np.nan to the onsets and offsets list and don't do any calculation for this peak.

The problem is that by adding np.nans to the arrays, it forces the array to be a float array instead of int, which means that before using the indices for slicing you have to convert them to int. Another option is to convert the arrays into object arrays )which I did) but idk if it is computationally optimal.

Additional changes:
- Removed unnecessary for loop in `_peaks_delineator`
- Add a np.nan in the onsets and offsets list if len(wt_peaks) == 0 instead of using the continue statement in `_onset_offset_delineator`. In this way the length of the arrays will be of the same length as R peaks which in my opinion makes easier further analysis. Additionally, before you discard the option of looking for an offset if for the onset `len(wt_peaks) == 0` which doesn't necessarily mean that you won't find an offset. 
- Code in a similar way  `_dwt_delineate_tp_onsets_offsets` and `_dwt_delineate_qrs_bounds` to make the code clearer.

Feel free to discard/adapt any of the changes I've implemented.

# Checklist

Here are some things to check before creating the PR.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -27,6 +27,7 @@ Contributors
 * `Robert Richer <https://github.com/richrobe>`_ *(FAU Erlangen-Nürnberg, Germany)*
 * `Russell Anderson <https://github.com/rpanderson>`_ *(La Trobe Institute for Molecular Science, Australia)*
 * `Alexander Wong <https://github.com/awwong1>`_ *(University of Alberta, Canada)*
+* `Raimon Padrós <https://github.com/raimonpv>`_ *(Universitat politècnica de Catalunya, Spain - Massachusetts General Hospital, USA)*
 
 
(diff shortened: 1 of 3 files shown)


DominiqueMakowski:
Awesome, thanks @raimonpv! also please consider adding yourself to the [contributors](https://github.com/neuropsychology/NeuroKit/blob/dev/AUTHORS.rst#contributors) ☺️

Tam-Pham:
Thanks @raimonpv for the improvements, especially for the `cwt method`. I have no objection to your updates. Please add yourself as contributor and we can merge 😄

Issue #316, opened 2020-07-07
Use Warning Control instead of `print("NeuroKit warning:")`


awwong1:
**Describe the solution you'd like**
Currently, neurokit warnings are printed using the `print` builtin function.
This should instead be replaced with calls to `warnings.warn`.

**How could we do it?**

Use [warnings.warn](https://docs.python.org/3/library/warnings.html#warnings.warn), follow [PEP 230](https://www.python.org/dev/peps/pep-0230/) for guidelines.

Example, stub `main.py` file.

```python
import warnings

class NeuroKitWarning(Warning): # UserWarning, DeprecationWarning, SyntaxWarning, RuntimeWarning
    pass

if __name__ == "__main__":
    freq = 234
    sampling_rate = 345
    nyquist = 456

    warnings.warn(
        f"Skipping requested frequency" +
        f" of {freq} Hz since it cannot be resolved at the" +
        f" sampling rate of {sampling_rate} Hz. Please increase" +
        f" sampling rate to {freq * 10} Hz or choose frequencies" +
        f" smaller than or equal to {nyquist} Hz.",
        category=NeuroKitWarning)

```
```text
main.py:17: NeuroKitWarning: Skipping requested frequency of 234 Hz since it cannot be resolved at the sampling rate of 345 Hz. Please increase sampling rate to 2340 Hz or choose frequencies smaller than or equal to 456 Hz.
  category=NeuroKitWarning)
```

**Why should we do it?**

- User control over printed warnings (suppressing warnings)
- Ability to filter out warnings that should be treated as errors

----

I would be happy to make this PR if this is something that other people are also interested in.

DominiqueMakowski:
Great idea, it's true that printing warnings does not seem to be the cleanest way 😁 

The only thing that I'd like, if possible, is to have the warning to include that it's a neurokit warning (in the message, so it stands out that it's from our code and not related to some other package or python) and also the function where the warning is located (for us to easily retrieve and debug). This information could be contained in the tracelog, but it'd be nice to have it explicitly in the warning message so it's eventually easy to report, fix or debug.

So please open a PR with like maybe one example and then we can see how it looks like, and if it works we go ahead to improve it throughout

PR #320, opened 2020-07-08
[Improvement] Refactor all printed NeuroKit warnings into warning control

awwong1:
# Description

This PR shows how to migrate the `print` calls to `warnings.warn`.
- https://github.com/neuropsychology/NeuroKit/issues/316
- https://docs.python.org/3/library/warnings.html#warnings.warn
- https://docs.pytest.org/en/latest/warnings.html#recwarn

# Proposed Changes

I replaced one warning in `signal_psd`, @DominiqueMakowski
> open a PR with like maybe one example and then we can see how it looks like, and if it works we go ahead to improve it throughout

Here's how it'll work if you call the function:

```python
import neurokit2 as nk

data = nk.data("bio_eventrelated_100hz")

out = nk.signal_psd(data["ECG"], sampling_rate=100)
```
```text
/home/alexander/sandbox/src/github.com/awwong1/NeuroKit/neurokit2/signal/signal_psd.py:112: NeuroKitWarning: signal_psd(): The duration of recording is too short to support a sufficiently long window for high frequency resolution. Consider using a longer recording or increasing the `min_frequency`
  category=NeuroKitWarning
```

# Checklist

Here are some things to check before creating the PR.


DominiqueMakowski:
Looks good, I'd say we schedule this improvement for the next release (as we'll probably merge the current one today or asap).

From that example you mentioned it seems indeed as repeating the function is not necessarily as it shows the line and the file... what do you think?

Also, is it possible to wrap this:
```python
warn(
                "signal_psd(): The duration of recording is too short to support a "
                "sufficiently long window for high frequency resolution. Consider using a longer recording "
                "or increasing the `min_frequency`"
                "or increasing the `min_frequency`",
                category=NeuroKitWarning
            )
```

in a function like `throw_warning(message)`? I'm not sure about the benefits other than having not to write the category arg and also having not to import everytime the warnings module (tho we would still need to import the warnings where `throw_warning` would be defined, so it's not a big gain)

awwong1:
The issue of wrapping the `warnings.warn` call in another function is the warning line number that appears will not reflect the location of the actual warning.

e.g. if `misc/warnings.py` had:
```python3
# -*- coding: utf-8 -*-
from warnings import warn

class NeuroKitWarning(RuntimeWarning):
    """Category for runtime warnings that occur within the NeuroKit library.
    """


def throw_warning(message):
    warn(message, category=NeuroKitWarning)
```

calling `throw_warning('whatever')` will show the error coming from `...neurokit2/misc/warnings.py:10:` no matter where you call the wrapper function.

Would you like me to change the rest of the print warnings? I will not modify any other neurokit warnings (such as the ImportErrors).

DominiqueMakowski:
> The issue of wrapping the warnings.warn call in another function is the warning line number that appears will not reflect the location of the actual warning.

Ah right yes! in that case indeed doesn't make sense to wrap it indeed.

> Would you like me to change the rest of the print warnings? I will not modify any other neurokit warnings (such as the ImportErrors).

Yes I think we can go ahead and do that, with removing the function name from the warning message (as it's redundant). Thanks!

DominiqueMakowski:
@awwong1 do you want me to merge this PR so we can open a new one with more or keep pushing to this one?

awwong1:
Let's keep this PR open, I plan on having the changes finished by the end of day.

awwong1:
> @awwong1 Thanks so much for your improvements, especially the incorporation of pytest warnings in our test files! Just picked up on these two errors and once these are changed we can merge smile

I am grateful that you took the time to review my changes. Thank you for humoring my print statement nitpicking. :smile_cat:

neurokit2/eog/eog_plot.py
```diff
@@ -49,7 +49,10 @@ def eog_plot(eog_signals, peaks=None, sampling_rate=None):
 
     # Sanity-check input.
     if not isinstance(eog_signals, pd.DataFrame):
-        print("NeuroKit error: The `eog_signals` argument must be the DataFrame returned by `eog_process()`.")
+        raise ValueError(
```
zen-juen, line 52: can add '*eog_plot()*' in "NeuroKit error: eog_plot(): The `eog_signals` ...."

awwong1: Good catch, will modify + add test

PR #327, opened 2020-07-10
[feature] Time-frequency methods

Tam-Pham:
- [x] short-time fourier transform

- [x] continuous wavelet transform

- [x] wigner ville distribution

- [x] pseudo wigner ville distribution


DominiqueMakowski:
I think we could merge that and fix the smooth pseudo wigner ville distribution later no?

DominiqueMakowski:
the README examples seems to fail in the latest commit, did we modify anything that could have changed its behaviour or is it a false positive?

PR #331, opened 2020-07-13
[Feature] EEG microstates analysis (basics)

DominiqueMakowski:
TODO list:

- [ ] Check current mod k-means against https://github.com/Frederic-vW/eeg_microstates/blob/master/eeg_microstates.py#L518
- [x] Add more clustering methods
- [ ] Add `microstates_findnumber()`
- [ ] Compute fit quality for each datapoint
- [ ] improving the feature extraction (Markov chains better statistics)
- [ ] Check peak detection, 'cause most of the publications find the GFP peaks, BUT for instance [here](https://www.researchgate.net/publication/7432398_Response_inhibition_deficits_in_externalizing_child_psychiatric_disorders_An_ERP-study_with_the_Stop-task) it says *"Microstate boarders were determined by relative **minima of GFP** together with relative **maxima in Diss**."*
- [x] Invent some way to automatically classify / sort the microstates, maybe depending on parameters like axis of symmetry or something 🤷


neurokit2/stats/cluster.py
```diff
@@ -375,6 +375,136 @@ def _cluster_mixture(data, n_clusters=2, bayesian=False, random_state=None, **kw
...
+        """
+        dx = np.diff(x) # discrete 1st derivative
+        zc = np.diff(np.sign(dx)) # zero-crossings of dx
+        m = 1 + np.where(zc == -2)[0] # indices of local max.
+        return m
+
+    #print("\n\t--- AAHC ---")
+    nt, nch = data.shape
+
+    # --- get GFP peaks ---
```
DominiqueMakowski, line 417: this should not be done in the clustering since its preprocesing

neurokit2/stats/cluster.py
```diff
@@ -375,6 +375,136 @@ def _cluster_mixture(data, n_clusters=2, bayesian=False, random_state=None, **kw
...
+    gfp = data.std(axis=1)
+    gfp_peaks = locmax(gfp)
+    #gfp_peaks = gfp_peaks[:100]
+    #n_gfp = gfp_peaks.shape[0]
+    gfp2 = np.sum(gfp**2) # normalizing constant in GEV
+
+    # --- initialize clusters ---
+    maps = data[gfp_peaks,:]
+    # --- store original gfp peaks and indices ---
+    cluster_data = data[gfp_peaks,:]
```
DominiqueMakowski, line 427: this is the data that this function should start with

neurokit2/stats/cluster.py
zen-juen: The algorithm seems like it requires the data (before and after peak extraction) as well as the calculation of gfp to even do any clustering in the first place though

DominiqueMakowski:
@Tam-Pham @zen-juen  Things to double-check / solve:

- [ ] cross validation index: https://github.com/Frederic-vW/eeg_microstates/blob/master/eeg_microstates.py#L594

Tried to implement it in clusters_quality but gives funny results. (especially the normalization part doesn't work).

- [ ] GEV

Implementations from https://github.com/Frederic-vW/eeg_microstates/blob/master/eeg_microstates.py#L602 and https://github.com/wmvanvliet/mne_microstates/blob/master/microstates.py#L126 (added my version of both in clusters_quality) give very different results. Also, it would be good to get the GEV per microstate, which currently isn't computed it seems.

- [ ] Performance 

Improve performance if possible.

- [ ] In kmod microstates corresponding to each datapoint are assigned using some `np.dot` magic that I don't understand. What is even weirder is that if we run a regular kmeans from sklearn and we take its predicted clusters and centroiods, and that we apply the same procedure as for modified kmeans (i.e., using np.dot), it gives different (and better ?!!!) predicted clusters... I don't understand it.


One of the main critical point of caution is the shape of the data, because the eeg data is by default in a wide format for mne and thus mne_microstates, but the clusters are in "long" format, whereas in eeg_micrsotates by Frederic, as well as our clustering algorithm, data is expected to be long format (rows and columns). This is why there is a lot of matrix transposition and possible errors resulting from this.

Issue #312, opened 2020-07-07
Interesting ideas and algorithms from the Bio-SP tool (matlab toolbox) by Nabian  et al. (2018)


DominiqueMakowski:
Very interesting open-access paper and toolbox by [Nabian et al. (2018)](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6231905/), with a similar scope to NK (but for [matlab](https://www.mathworks.com/matlabcentral/fileexchange/64013-biosignal-specific-processing-bio-sp-tool)). The paper is very well written and the algorithms and methods are clearly described. It contains some interesting features ideas and algorithms that we could maybe implement to facilitate cross-language comparisons.

If a given method is specific to this package (no prior publication), we can name it `*_function_nabian2018()` or `_function_biosp()`, and of course additionally reference and link this paper (and the original reference if applicable).

### ECG

- [ ] R peak detection

> Step I: Since the human heart rate does not exceed 5 beats per second two adjacent QRS complexes cannot appear closer together than 200ms [14]. Therefore, a sliding window of 400 ms size with a step size of 1 sample can iteratively scan the entire signal. At each iteration, we find the global maximum in the sliding window and if this global maximum appears in the middle of the window, that global maximum is marked as a potential R point. By scanning the entire input ECG signal, the peaks which are likely to be R peaks are detected. The ith peak is denoted by Ri.

- [ ] R peak fix

> Step II: Ri peaks that are lower in amplitude than an amplitude threshold (Amp-Thr) are eliminated. Amp-Thr is initially set to 1/3 times the maximum amplitude of the first 2 seconds of the input ECG signal. To adapt to the changes in the ECG over time, upon passing sufficient R peaks from the beginning of the signal, Amp-Thr is updated to 0.75 times the mean of the amplitude of the last eight determined R peaks.

> Step III: After scanning through all the detected adjacent R peak to R peak time intervals (R-R intervals), if any R-R interval is greater than a predefined threshold (RR-Thr), it implies that an R peak is missing. Thus, we assign a new R peak as the global maximum amplitude of the signal within the range of Ri+200ms and Ri+1-200ms [14]. According to [14], the rate of change for adjacent R-R intervals will not exceed 166%. Thus, RR-Thr is initially set to the previous R-R interval multiplied by 1.66. Upon detecting enough R-R intervals, the RR-Thr is then updated to reflect the average of the last eight R-R intervals multiplied by 1.66.

- [ ] QRS delineation

[...]

PR #339, opened 2020-07-22
[Features] PPG preprocessing 

elaineteo2000:


neurokit2/ppg/ppg_clean.py
```diff
@@ -33,27 +35,62 @@ def ppg_clean(ppg_signal, sampling_rate=1000, method="elgendi"):
...
 
 def _ppg_clean_elgendi(ppg_signal, sampling_rate):
 
     filtered = signal_filter(
         ppg_signal, sampling_rate=sampling_rate, lowcut=0.5, highcut=8, order=3, method="butter_ba"
     )
     return filtered
+
+
+def _ppg_clean_nabian2018(ppg_signal, sampling_rate, heart_rate=None):
```
zen-juen, line 79: @elaineteo2000 Great attempt on your first PR Elaine! 😄 

I've reformatted the function to have:
1) ``highcut`` not as an argument since the general API of our _clean* functions do not require users to specify any filter related info, so we will just set highcut=40 in the function itself for now

2) Initially you specified that if highcut is not default (40Hz in this case), then it would be at least 10 times the heart rate and less than 0.5times the sampling rate. However ``highcut`` needs a specific value and not a range which is what I think you were trying to do? 😄 So what I've done here is just to check that the default value of 40Hz is meets these above criteria since ``highcut=40`` is the unchangeable, default cut off for now!

3) Great use of ``nk.signal_filter()``! 👍 I put ``highcut=highcut`` here because I have saved (line 84) ``highcut=40`` , initially I think you put ``highcut='default'`` and remember the ``signal_filter()`` function won't be able to process this argument because it is a string and not a number!

zen-juen:
I think there are more PPG features that can be added based on https://github.com/neuropsychology/NeuroKit/issues/312 so let's leave this PR open for now!

DominiqueMakowski:
Nah, I think we close it and we open new ones for new features. We need to try to keep the PRs smol 😁

Issue #243, opened 2020-06-06
signal_interpolate(): dealing with physiologically implausible values caused by under- or overshooting interpolation


DominiqueMakowski:
Following #180 and #242 interpolation can create crazy values, extending far beyond the original range. 

We could add a checker for that, throwing a warning for instance if the new range is extended by more than 25% (or anything else): `"Warning: the " +  interpolation_method + " method for interpolation created some implausible values. Proceed with caution, visualize the resulting interpolated signal and consider switching to linear interpolation."`

Additionally, we could add a `safe` or `cap` argument that would, for instance, replace all interpolated values outside this "safe" zone (e.g., 125% of range) by either the max/min or by NaNs?

@JanCBrammer thoughts?

JanCBrammer:
@DominiqueMakowski, I noticed this behavior of the interpolation while implementing `ppg_simulate()`. I had difficulty finding words for this behavior, but now that it comes up again I did some research. The behavior we see is referred to a "overshooting" or "undershooting" of the interpolated values in the y direction. This happens when the interpolation method is not monotone.

These three resources help explain and visualize this.

* https://www.orcina.com/webhelp/OrcaFlex/Content/html/Interpolationmethods.htm
* https://en.wikipedia.org/wiki/Monotone_cubic_interpolation
* http://articles.adsabs.harvard.edu/cgi-bin/nph-iarticle_query?1990A%26A...239..443S&defaultprint=YES&filetype=.pdf

The solution is to use methods that preserve monotonocity (i.e., that do not overshoot or undershoot the y values). These methods are linear interpolation, [Akima interpolation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.Akima1DInterpolator.html#scipy.interpolate.Akima1DInterpolator) (I use this one in `ppg_simulate()`), or [Monotone cubic spline interpolation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.PchipInterpolator.html#scipy.interpolate.PchipInterpolator). I suggest we use these methods instead of trying to correct or warn against over- / undeshooting. The latter could unnecessarily confuse users and clutters the code.

DominiqueMakowski:
Thanks for the resources!
Do you think we should replace the cubic default by one of these then? Or do they, on average, not perform better? (the goal of the default method being to have something that works the best on average, in the majority of cases and that is not too dangerous)

JanCBrammer:
[...]

PR #341, opened 2020-07-23
[Feature] Add new respiration rate methods

Tam-Pham:


neurokit2/rsp/rsp_rate.py
```diff
@@ -0,0 +1,130 @@
...
+    rsp = signal_resample(rsp_cleaned, sampling_rate=sampling_rate, desired_sampling_rate=desired_sampling_rate)
+
+    # Define paramters
+    window_length = int(desired_sampling_rate * window)
+
+    rsp_rate = []
+    for start in np.arange(0, N, hop_size):
+        window_segment = rsp[start: start + window_length]
+        if len(window_segment) < window_length:
+            continue # the last frames that are smaller than windlow_length
```
JanCBrammer, line 80: maybe `break` instead of `continue` (no need to iterate over the rest)?

neurokit2/rsp/rsp_rate.py
```diff
@@ -0,0 +1,130 @@
...
+    >>> rsp_rate = nk.rsp_rate(rsp_cleaned, method="peaks")
+    >>> rsp_rate2 = nk.rsp_rate(rsp_cleaned, method="xcorr")
+    """
+
+    if method.lower() in ["peak", "peaks", "signal_rate"]:
+        if peaks is None:
+            peaks, info = rsp_peaks(rsp_cleaned, sampling_rate=sampling_rate, method=peak_method)
+        rsp_rate = signal_rate(peaks, sampling_rate=sampling_rate, desired_length=len(rsp_cleaned))
+
+    if method.lower() in ["cross-correlation", "xcorr"]:
```
DominiqueMakowski, line 53: I'd also probably do an if, elif, else error "wrong method" flow to throw informative errors

neurokit2/rsp/rsp_rate.py
```diff
@@ -38,22 +35,28 @@ def rsp_rate(rsp_cleaned, peaks=None, sampling_rate=1000, window=10, hop_size=1,
...
-    >>> rsp_rate2 = nk.rsp_rate(rsp_cleaned, method="xcorr")
+    >>> rsp_rate_peak = nk.rsp_rate(rsp_cleaned, sampling_rate=sampling_rate, method="peaks")
+    >>> rsp_rate_xcorr = nk.rsp_rate(rsp_cleaned, sampling_rate=sampling_rate, method="xcorr")
     """
 
     if method.lower() in ["peak", "peaks", "signal_rate"]:
         if peaks is None:
             peaks, info = rsp_peaks(rsp_cleaned, sampling_rate=sampling_rate, method=peak_method)
-        rsp_rate = signal_rate(peaks, sampling_rate=sampling_rate, desired_length=len(rsp_cleaned))
+        rsp_rate = signal_rate(peaks, sampling_rate=sampling_rate, desired_length=len(rsp_cleaned), interpolation_method="linear")
```
DominiqueMakowski, line 48: I wouldn't hardcode the interpolation_method nor put linear as a default, I think we discussed it in https://github.com/neuropsychology/NeuroKit/issues/243 and opted for monotonic if I remember correctly

Tam-Pham: Yes, you are right. I added in for comparison and didn't remove. I will add interpolation method as an option

PR #349, opened 2020-08-06
[Fix] Review microstate implementations

Tam-Pham:
Microstates list:

- [x] Check current mod k-means against https://github.com/Frederic-vW/eeg_microstates/blob/master/eeg_microstates.py#L518
- [x] Check other methods (ICA, PCA, AAHC etc.)

- [ ] Check peak detection, 'cause most of the publications find the GFP peaks, BUT for instance [here](https://www.researchgate.net/publication/7432398_Response_inhibition_deficits_in_externalizing_child_psychiatric_disorders_An_ERP-study_with_the_Stop-task) it says *"Microstate boarders were determined by relative **minima of GFP** together with relative **maxima in Diss**."*
- [x] cross validation index: https://github.com/Frederic-vW/eeg_microstates/blob/master/eeg_microstates.py#L594

Tried to implement it in clusters_quality but gives funny results. (especially the normalization part doesn't work).

- [x] GEV

Implementations from https://github.com/Frederic-vW/eeg_microstates/blob/master/eeg_microstates.py#L602 and https://github.com/wmvanvliet/mne_microstates/blob/master/microstates.py#L126 (added my version of both in clusters_quality) give very different results. Also, it would be good to get the GEV per microstate, which currently isn't computed it seems.

### DIFF
--- a/neurokit2/stats/cluster_quality.py
+++ b/neurokit2/stats/cluster_quality.py
@@ -86,7 +86,7 @@ def cluster_quality(data, clustering, clusters=None, info=None, n_random=10, **k
 
     # Variance explained
     general["Score_VarianceExplained"] = _cluster_quality_variance(data, clusters)
-    general["Score_GEV"] = _cluster_quality_gev(data, clusters, clustering, **kwargs)
+    general["Score_GEV"], _ = _cluster_quality_gev(data, clusters, clustering, **kwargs)
     general["Score_CrossValidation"] = _cluster_quality_crossvalidation(data, clusters, clustering)
 
     # Gap statistic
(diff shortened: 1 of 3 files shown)


neurokit2/stats/cluster_quality.py
```diff
@@ -86,7 +86,7 @@ def cluster_quality(data, clustering, clusters=None, info=None, n_random=10, **k
 
     # Variance explained
     general["Score_VarianceExplained"] = _cluster_quality_variance(data, clusters)
-    general["Score_GEV"] = _cluster_quality_gev(data, clusters, clustering, **kwargs)
+    general["Score_GEV"], _ = _cluster_quality_gev(data, clusters, clustering, **kwargs)
```
DominiqueMakowski, line 89: we should add individual GEVs in the output of cluster_quality

Tam-Pham: Currently it's returned in the info of microstates_segment(). I'm not sure how should we return it in cluster_quality()

DominiqueMakowski: ok let's leave it as is for now

PR #383, opened 2020-11-16
Add hrstd argument to ecg_simulate

jrepifano:
Added argument to ecg_simualate to allow users to specify hr std when generating with ecgsyn method.

# Description

This PR aims to allow users to specify hr std to mimic the functionality of the original ecgsyn implementaiton.

# Proposed Changes

I added the argument for hr std in the ecg_simulate function and kept 1 the default in the case that it is not specified.

### DIFF
--- a/neurokit2/ecg/ecg_simulate.py
+++ b/neurokit2/ecg/ecg_simulate.py
@@ -8,7 +8,7 @@ from ..signal import signal_distort, signal_resample
 
 
 def ecg_simulate(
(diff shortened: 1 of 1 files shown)


Tam-Pham:
hi @jrepifano 

Thank you for your PR. I think it's a great add-on. We will merge it to dev and the option will be available for users in the next version :)

DominiqueMakowski:
@jrepifano I forgot to add, but feel free to open a new PR to add yourself in the [contributors list](https://github.com/neuropsychology/NeuroKit/blob/master/AUTHORS.rst#contributors)

PR #389, opened 2020-11-26
0.1.0 🎉 

DominiqueMakowski:
- [x] Add reference to publication once we have it


DominiqueMakowski:
I wanted to wait 'til the paper is "out" (or at least, when we have a preliminary reference) for merging this, but I have received no news from production yet, possibly related to the hollidays and all. So we'll probably need to merge this before that, as a substantial number of small fixes are in, and we'll update the citation once we have it.

By the way, a (late) merry Christmas and happy holidays to all 🎉 🤗 looking forward to next year!

PR #398, opened 2020-12-26
[Fix] Fix offset bug in neurokit2/ecg/ecg_delineate.py

jluby:
# Description

This PR aims to fix a small bug in neurokit2/ecg/ecg_delineate.py which arises because the emptiness check of 'candidate_offsets' is performed using 'onset_slope_peaks' instead of 'offset_slope_peaks'. This PR prevents the function from erroneously passing the emptiness check and subsequently throwing an error when attempting to get 'candidate_offsets[0]'. 

# Proposed Changes

The only proposed change is correcting line 368's 'onset_slope_peaks' reference to 'offset_slope_peaks'.


Tam-Pham:
Hi @jluby 

Thank you for opening a PR with us. Upon reviewing #397, I realized the existence of this bug too 😄

PR #399, opened 2020-12-26
[Fix] Fix signal_zerocrossings indexing issues

jluby:
# Description

Throughout the package, signal_zerocrossings is used to find indices where signal crosses 0. Before calling signal_zerocrossings, signals are, correctly, checked to ensure that peaks[current_idx] is positive and peaks[next_idx] is negative. However, peaks[current_idx:next_idx] is used in the signal_zerocrossings call, excluding peaks[next_idx] value. When signal is positive until x[next_idx], the function currently throws error.

# Proposed Changes

All signal_zerocrossings calls are changed so that the second peak is included in signal_zerocrossings calls. That is, peaks[current_idx:next_idx] becomes peaks[current_idx:next_idx+1] in all cases.


DominiqueMakowski:
Looks good, thanks a lot!

Issue #407, opened 2021-01-25
ECG signal quality classifier by Zhao (2018)


DominiqueMakowski:
- [Zhao, Z., & Zhang, Y. (2018). "SQI quality evaluation mechanism of single-lead ECG signal based on simple heuristic fusion and fuzzy comprehensive evaluation". Frontiers in Physiology, 9, 727.](https://www.frontiersin.org/articles/10.3389/fphys.2018.00727/full)
- Current implementation in biosppy by @TiagoTostas: PIA-Group/BioSPPy#75

> This SQI takes makes use of four different quality indexes. Each one of the four indexes is classified as either optimal, suspicious or unqualified and these are then used to classify the signal according to the following function:
<img width="469" alt="Screenshot 2021-01-24 at 21 07 44" src="https://user-images.githubusercontent.com/39189659/105642151-428c8280-5e88-11eb-95d4-75d02e32b75e.png">

> Using one detector makes it impossible to extract one of the features, the matching degree between the QRS, sSQI. You could adjust the decision rule and use only the other indexes three indexes (power of baseline, power of QRS and kurtosis of the signal). Any particular reason for that?

_Originally posted by @TiagoTostas in https://github.com/neuropsychology/NeuroKit/issues/124#issuecomment-766424249_

DominiqueMakowski:
Current implementation in biosppy by @TiagoTostas: https://github.com/PIA-Group/BioSPPy/pull/75

DominiqueMakowski:
It is a threshold classifier based on several indices (referred to as Signal Quality Indexes - SQI):

- Matching Degree of R Peak Detection (qSQI): The corespodnance of two different R-detection methods, in their case: *Hilbert and dynamic adaptive threshold based on R-wave detection* and *wavelet transform*.

> for wavelet transform, the input ECG signal was decomposed by discrete wavelet transform (DWT) with four layers, and the wavelet coefficients for each layer were obtained. The R point is the singularity of the ECG waveform. The extraction of the R point is completed by the correspondence between the singularity of the signal and the positive and negative modulus maximum of the wavelet coefficients to the zero point (Zhen et al., 2008)

- Power Spectrum Distribution of QRS Wave (pSQI)
- Variability in the R-R Interval (cSQI) (apparently redundant)
- Skewness sSQI and Kurtosis (kSQI)
- The Relative Power in the Baseline (basSQI)

[...]

PR #410, opened 2021-02-08
[feature] ecg quality method by zhao2018

DominiqueMakowski:
- See https://github.com/neuropsychology/NeuroKit/issues/407 for rationale and reference

Goal is to add a ECG quality classifier based on these metrics:

- [x] Power Spectrum Distribution of QRS Wave (pSQI)
- [x] Kurtosis (kSQI)
- [x] Relative Power in the Baseline (basSQI)

### DIFF
--- a/neurokit2/ecg/ecg_quality.py
+++ b/neurokit2/ecg/ecg_quality.py
@@ -8,13 +8,15 @@ from .ecg_peaks import ecg_peaks
 from .ecg_segment import ecg_segment
 
 
-def ecg_quality(ecg_cleaned, rpeaks=None, sampling_rate=1000):
+def ecg_quality(ecg_cleaned, rpeaks=None, sampling_rate=1000, method="averageQRS"):
     """Quality of ECG Signal.
 
-    Compute a continuous index of quality of the ECG signal, by interpolating the distance
+    The "averageQRS" method computes a continuous index of quality of the ECG signal, by interpolating the distance
     of each QRS segment from the average QRS segment present in the data. This index is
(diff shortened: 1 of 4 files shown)


neurokit2/ecg/ecg_quality.py
```diff
@@ -6,16 +6,18 @@
 from ..stats import distance, rescale
 from .ecg_peaks import ecg_peaks
 from .ecg_segment import ecg_segment
+from scipy import stats, integrate, signal
```
DominiqueMakowski, line 9: @maxyongg I prefer to have `import scipy` and then call explicitly `scipy.stats.kurtosis`, `scipy.signal.whatezer` etc.

DominiqueMakowski:
@zen-juen could you take a look at that I also said to @maxyongg to give it a look as this is a feateure I want to add

zen-juen:
@maxyongg @DominiqueMakowski I've refactored the code structure so it abides more by nk's naming conventions + only included the internal functions of pSQI, kSQI and basSQI (leaving out the qSQI)

However, one thing to note is that in zhao et al's "fuzzy" method, the indices are given different weights to determine the eventual signal quality classification, with:
```
W = np.array([0.4, 0.4, 0.1, 0.1])
```
respectively for qSQI, pSQI, kSQI and basSQI - and according to zhao, this weighting results in an accurate quality assessment. Because we left out qSQI, we'll have to rearrange the weights 🤔 for now, I've set it arbitrarily to `W = np.array([0.6, 0.2, 0.2])`

PR #423, opened 2021-03-17
Optimize _ecg_findpeaks_MWA

jukka:
This optimization speeds up the moving average computation by an order of
magnitude. This is the most time-consuming part of pantompkins peak
detection, which after this optimization is about 5x faster on my computer.

Note the TODO about the _ecg_findpeaks_MWA output. I intentionally kept
this as a pure optimization with no impact on the output of the function,
but I do wonder whether the earlier implementation was actually incorrect,
or if the output is intentionally shifted by one.

# Checklist

### DIFF
--- a/neurokit2/ecg/ecg_findpeaks.py
+++ b/neurokit2/ecg/ecg_findpeaks.py
@@ -1006,29 +1006,28 @@ def _ecg_findpeaks_rodrigues(signal, sampling_rate=1000):
 
 
 def _ecg_findpeaks_MWA(signal, window_size):
-    """From https://github.com/berndporr/py-ecg-detectors/"""
+    """Based on https://github.com/berndporr/py-ecg-detectors/
 
-    mwa = np.zeros(len(signal))
+    Optimized for vectorized computation.
+    """
+
     sums = np.cumsum(signal)
 
-    def get_mean(begin, end):
-        if begin == 0:
-            return sums[end - 1] / end
+    # Compute moving average for the first `window_size` elements.
+    # The denominator grows until it reaches `window_size`.
+    mwa_head = sums[:window_size] / np.arange(1, window_size + 1)
 
-        dif = sums[end - 1] - sums[begin - 1]
-        return dif / (end - begin)
+    # Compute moving average for all the remaining elements based on the
+    # difference of the cumulative sum across `window_size` elements.
+    mwa_tail = (sums[window_size:] - sums[:-window_size]) / window_size
 
-    for i in range(len(signal)):  # pylint: disable=C0200
-        if i < window_size:
-            section = signal[0:i]
-        else:
-            section = get_mean(i - window_size, i)
-
-        if i != 0:
-            mwa[i] = np.mean(section)
-        else:
-            mwa[i] = signal[i]
-
+    # TODO: Is the following construct intentional? This produces exactly
+    # equal output as the code before the above optimization, but the shift
+    # by one seems like a potential bug in the earlier code. See the test
+    # case in tests/tests_ecg_findpeaks_MWA.py for example output.
+    mwa = np.concatenate([signal[0:1], mwa_head, mwa_tail[:-1]])
+    # Here's a possibly more correct version:
+    # mwas = np.concatenate([mwa_head, mwa_tail])
     return mwa
 
 
--- a/neurokit2/ecg/ecg_findpeaks.py
+++ b/neurokit2/ecg/ecg_findpeaks.py
@@ -1021,14 +1021,7 @@ def _ecg_findpeaks_MWA(signal, window_size):
     # difference of the cumulative sum across `window_size` elements.
     mwa_tail = (sums[window_size:] - sums[:-window_size]) / window_size
 
-    # TODO: Is the following construct intentional? This produces exactly
-    # equal output as the code before the above optimization, but the shift
-    # by one seems like a potential bug in the earlier code. See the test
-    # case in tests/tests_ecg_findpeaks_MWA.py for example output.
-    mwa = np.concatenate([signal[0:1], mwa_head, mwa_tail[:-1]])
-    # Here's a possibly more correct version:
-    # mwas = np.concatenate([mwa_head, mwa_tail])
-    return mwa
+    return np.concatenate([mwa_head, mwa_tail])
 
 
 def _ecg_findpeaks_peakdetect(detection, sampling_rate=1000):
--- a/tests/tests_ecg_findpeaks_MWA.py
+++ b/tests/tests_ecg_findpeaks_MWA.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+import numpy as np
+import pytest
+
+# Trick to directly access the internal function.
+# Using neurokit2.ecg.ecg_findpeaks._ecg_findpeaks_MWA doesn't
+# work because of the "from .ecg_findpeaks import ecg_findpeaks"
+# statement in neurokit2/ecg/__init.__.py.
+from neurokit2.ecg.ecg_findpeaks import _ecg_findpeaks_MWA
+
+
+# TODO: Is the expected output asserted by this test correct?
+# See the comment in _ecg_findpeaks_MWA for more details.
+def test_ecg_findpeaks_MWA():
+    np.testing.assert_array_equal(
+        _ecg_findpeaks_MWA(
+            np.array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9], dtype=np.float),
+            3),
+        [0, 0, 0.5, 1, 2, 3, 4, 5, 6, 7])
--- a/tests/tests_ecg_findpeaks_MWA.py
+++ b/tests/tests_ecg_findpeaks_MWA.py
@@ -9,11 +9,9 @@ import pytest
 from neurokit2.ecg.ecg_findpeaks import _ecg_findpeaks_MWA
 
 
-# TODO: Is the expected output asserted by this test correct?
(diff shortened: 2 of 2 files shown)


DominiqueMakowski:
Awesome, thanks a lot @jukka 👌👌 

> but I do wonder whether the earlier implementation was actually incorrect, or if the output is intentionally shifted by one.

What do you mean, do you suspect a bug? I think the original code was inspired from [here](https://github.com/berndporr/py-ecg-detectors/blob/master/ecgdetectors.py#L550-L563), but maybe we made an error during the inspiration?

jukka:
Looking at the code I do believe there's an off-by-one error in the [current implementation][1]. This doesn't seem to be the case in the [original code][2]. In particular `mwa[1]` ends up being `np.mean(signal[0:1])` (i.e. `== signal[0]` since end index of the slice is exclusive) with the current code, while the original code would set it to `np.mean(input_array[0:2])`. The latter behavior seems more correct, though without a comment or a test case to clarify the intended API contract it's hard to be sure.

In this PR I'm just reproducing the current behavior, but I'd be happy to send a potential fix with some more detailed analysis if the above sounds correct.

[1]: https://github.com/neuropsychology/NeuroKit/blob/883bbb3461d1e40dc5f16a1f6eb2a7babb9f0758/neurokit2/ecg/ecg_findpeaks.py#L1008-L1032
[2]: https://github.com/berndporr/py-ecg-detectors/blob/master/ecgdetectors.py#L550-L563

JanCBrammer:
@jukka, thanks for spotting this. Indeed nice potential for optimization / speed-up by vectorization since `_ecg_findpeaks_MWA()` is called in two of the ECG detectors (Pan-Tomkins, Elgendi) we took over from https://github.com/berndporr/py-ecg-detectors/.

Note that we have our own moving average implementation: https://github.com/neuropsychology/NeuroKit/blob/883bbb3461d1e40dc5f16a1f6eb2a7babb9f0758/neurokit2/signal/signal_smooth.py#L9

Below is a visual comparison of the three implementations. I feel like the NeuroKit implementation produces the most reasonable result? @jukka, @DominiqueMakowski, what do you think?
demo_mwas

<details>
<summary>Script to reproduce the figure</summary>
<p>

```Python
import numpy as np
from neurokit2.signal import signal_smooth


def mva_original(signal, window_size):

    mwa = np.zeros(len(signal))
    sums = np.cumsum(signal)

    def get_mean(begin, end):
        if begin == 0:
            return sums[end - 1] / end

        dif = sums[end - 1] - sums[begin - 1]
        return dif / (end - begin)

    for i in range(len(signal)):
        if i < window_size:
            section = signal[0:i]
        else:
            section = get_mean(i - window_size, i)

        if i != 0:
            mwa[i] = np.mean(section)
        else:
            mwa[i] = signal[i]

    return mwa


def mva_Jukka(signal, window_size):

    sums = np.cumsum(signal)

    # Compute moving average for the first `window_size` elements.
    # The denominator grows until it reaches `window_size`.
    mwa_head = sums[:window_size] / np.arange(1, window_size + 1)

    # Compute moving average for all the remaining elements based on the
    # difference of the cumulative sum across `window_size` elements.
    mwa_tail = (sums[window_size:] - sums[:-window_size]) / window_size

    # TODO: Is the following construct intentional? This produces exactly
    # equal output as the code before the above optimization, but the shift
    # by one seems like a potential bug in the earlier code. See the test
    # case in tests/tests_ecg_findpeaks_MWA.py for example output.
    # mwa = np.concatenate([signal[0:1], mwa_head, mwa_tail[:-1]])
    # Here's a possibly more correct version:
    mwa = np.concatenate([mwa_head, mwa_tail])

    return mwa

window_size = 5
np.random.seed(42)
signal = np.random.random(50)
signal_smooth1 = mva_original(signal, window_size)
signal_smooth2 = mva_Jukka(signal, window_size)
signal_smooth3 = signal_smooth(signal, kernel="boxcar", size=window_size)

print(signal_smooth1.size, signal_smooth2.size, signal_smooth3.size)

fig, ax = plt.subplots(3, 1, sharex=True)
ax[0].plot(signal)
ax[0].plot(signal_smooth1)
ax[0].set_title("_ecg_findpeaks_MWA")
ax[1].plot(signal)
ax[1].plot(signal_smooth2)
ax[1].set_title("Jukka's refactor of _ecg_findpeaks_MWA")
ax[2].plot(signal)
ax[2].plot(signal_smooth3)
ax[2].set_title("NeuroKit's signal_smooth with boxcar kernel")

plt.show()

**Comment by jukka** [comment]
Neat! I agree that `signal_smooth` does best match the signal. The improvement is easy to see on a simulated ECG signal with 50Hz sampling:

ecg_mwa

My one concern here is performance. The reason why I looked into this optimization is that I have dozens of gigabytes of ECG data to process, and I was hitting performance issues with the current implementation. Unless `np.convolve` has a special case optimization for boxcar kernels, it's hard to see it performing as well as the `np.cumsum` based implementation. But I haven't measured the performance yet.

Also I wonder how much the presence or magnitude of a signal shift matters here. For my use case I only care about heart rate computed from the peaks, so a small constant shift in the signal doesn't make any difference. But others might rely more on that information.

**Comment by jukka** [comment]
Assuming I do find a performance loss with `np.convolve`, would it make sense to add this optimization to `signal_smooth` as a special case for "boxcar" kernels?

**Comment by DominiqueMakowski** [comment]
If I understood correclty, there are two separate things here, 1) an off-by-one error bug in the implementation of MWA and 2) possible performance improvement. So it seems like we can fix these two for this PR, and then in a new one eventually modify / simplify / expand the API?

**Comment by JanCBrammer** [comment]
> Assuming I do find a performance loss with `np.convolve`, would it make sense to add this optimization to `signal_smooth` as a special case for "boxcar" kernels?

Absolutely! If the `cumsum` implementation outperforms the `np.convolve` implementation, let's update `signal_smooth` accordingly.

**Comment by JanCBrammer** [comment]
> Also I wonder how much the presence or magnitude of a signal shift matters here. For my use case I only care about heart rate computed from the peaks, so a small constant shift in the signal doesn't make any difference. But others might rely more on that information.

I think for event-related analyses that require some temporal precision the shift might be problematic. Couldn't think of a case though where that much temporal precision would be required in ECG analyses.

**Comment by JanCBrammer** [comment]
@jukka, technically though any [causal filter](https://en.wikipedia.org/wiki/Causal_filter) would shift the smoothed signal, right? Not sure if causality is something that we want to consider / enforce.

**Comment by jukka** [comment]
The added commit 331a3f3995bb79efffb72344108000bf680e8803 fixes the off-by-one issue and makes the output match that of [MWA_original][1] in py-ecg_detectors:
ecg_mwa

Script to reproduce the figure:
```python
import numpy as np

def mva_original(signal, window_size):

    mwa = np.zeros(len(signal))
    sums = np.cumsum(signal)

    def get_mean(begin, end):
        if begin == 0:
            return sums[end - 1] / end

        dif = sums[end - 1] - sums[begin - 1]
        return dif / (end - begin)

    for i in range(len(signal)):
        if i < window_size:
            section = signal[0:i]
        else:
            section = get_mean(i - window_size, i)

        if i != 0:
            mwa[i] = np.mean(section)
        else:
            mwa[i] = signal[i]

    return mwa


def mva_Jukka(signal, window_size):

    sums = np.cumsum(signal)

    mwa_head = sums[:window_size] / np.arange(1, window_size + 1)

    mwa_tail = (sums[window_size:] - sums[:-window_size]) / window_size

    return np.concatenate([mwa_head, mwa_tail])


def ecgdetectors_MWA_original(input_array, window_size):

    mwa = np.zeros(len(input_array))
    mwa[0] = input_array[0]
    
    for i in range(2,len(input_array)+1):
        if i < window_size:
            section = input_array[0:i]
        else:
            section = input_array[i-window_size:i]        
        
        mwa[i-1] = np.mean(section)

    return mwa


window_size = 5
np.random.seed(42)
signal = np.random.random(50)

signal_smooth1 = mva_original(signal, window_size)
signal_smooth2 = mva_Jukka(signal, window_size)
signal_smooth3 = ecgdetectors_MWA_original(signal, window_size)

fig, ax = plt.subplots(3, 1, sharex=True)
ax[0].plot(signal)
ax[0].plot(signal_smooth1)
ax[0].set_title("_ecg_findpeaks_MWA")
ax[1].plot(signal)
ax[1].plot(signal_smooth2)
ax[1].set_title("Jukka's refactor of _ecg_findpeaks_MWA")
ax[2].plot(signal)
ax[2].plot(signal_smooth3)
ax[2].set_title("MWA_original in py-ecg-detectors")

plt.show()
```


[1]: https://github.com/berndporr/py-ecg-detectors/blob/81b271f4c6adcb1226a92ad61f09af8819768bf6/ecgdetectors.py#L550-L563

jukka:
Re: causal; I don't have enough expertise in this space (my background is in computer science, not bio signals) to have an informed opinion. I do note though that the Pan-Tompkins paper describes a real-time algorithm which by definition is causal.

JanCBrammer:
> [331a3f3](https://github.com/neuropsychology/NeuroKit/commit/331a3f3995bb79efffb72344108000bf680e8803) fixes the off-by-one issue and makes the output match that of [MWA_original](https://github.com/berndporr/py-ecg-detectors/blob/81b271f4c6adcb1226a92ad61f09af8819768bf6/ecgdetectors.py#L550-L563) in py-ecg_detectors

@jukka, thanks!

> Pan-Tompkins paper describes a real-time algorithm which by definition is causal

Exactly, I think it's good to stay as close to the original implementations as possible / preserve their real-time characteristics.

PR #427, opened 2021-03-19
0.1.2

Tam-Pham:


neurokit2/hrv/hrv_nonlinear.py
```diff
@@ -186,6 +192,9 @@ def hrv_nonlinear(peaks, sampling_rate=1000, show=False):
     # Entropy
     out["ApEn"] = entropy_approximate(rri, delay=1, dimension=2, r=0.2 * np.std(rri, ddof=1))
```
DominiqueMakowski, line 193: @zen-juen should compute r first and then pass it to avoid multiple computation [efficiency :p]

PR #450, opened 2021-05-10
Optimize _ecg_findpeaks_peakdetect

jukka:
# Description

I noticed a lot of time being spent in this function when processing a large corpus of ECG recordings. The reason is that the original code from py-ecg-detectors is mostly not vectorized and the parts that are vectorized do lots of unnecessary work.

# Proposed Changes

I broke this change into several smaller commits to make it easier to understand the optimizations and to verify that they don't accidentally change the output of peak detection. Additionally the first commit just introduces a change aversion test case as an additional safeguard against accidental changes in the output of this function. The test passes both before and after this optimization.

Note that while working on this change I noticed that _ecg_findpeaks_peakdetect does *not* detect flat peaks, i.e. local maxima with more than one subsequent sample having the exact same value. I didn't want to change that behavior without deeper understanding of potential downstream effects, but I did leave a TODO in case it's worth following up on separately.

A simple benchmark shows order-of-magnitude speedups as a result of this optimization.

Before:

```
-------------------------------------------------------------------------------------------- benchmark: 2 tests -------------------------------------------------------------------------------------------
Name (time in ms)                               Min                Max               Mean            StdDev             Median               IQR            Outliers      OPS            Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_ecg_findpeaks_peakdetect_good_4000     70.1957 (1.0)      72.1044 (1.0)      70.7616 (1.0)      0.5059 (1.05)     70.7356 (1.0)      0.4704 (1.0)           5;1  14.1319 (1.0)          15           1
test_ecg_findpeaks_peakdetect_bad_500       73.9374 (1.05)     75.4275 (1.05)     74.4368 (1.05)     0.4799 (1.0)      74.2421 (1.05)     0.7253 (1.54)          4;0  13.4342 (0.95)         14           1
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

After:

```
------------------------------------------------------------------------------------------ benchmark: 2 tests ------------------------------------------------------------------------------------------
Name (time in ms)                              Min               Max              Mean            StdDev            Median               IQR            Outliers       OPS            Rounds  Iterations
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_ecg_findpeaks_peakdetect_good_4000     2.7121 (1.0)      4.2319 (1.0)      2.8094 (1.0)      0.1001 (1.26)     2.7969 (1.0)      0.0464 (1.0)          8;12  355.9522 (1.0)         334           1
test_ecg_findpeaks_peakdetect_bad_500       3.3477 (1.23)     4.5243 (1.07)     3.5269 (1.26)     0.0793 (1.0)      3.5247 (1.26)     0.0767 (1.65)         32;3  283.5385 (0.80)        280           1
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

<details>
<summary>pytest-benchmark code</summary>

```python
import numpy as np
import os.path
import pandas as pd

from neurokit2.ecg.ecg_findpeaks import _ecg_findpeaks_peakdetect

def _read_csv_column(csv_name, column):
    csv_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), 'tests', 'ecg_data', csv_name)
    csv_data = pd.read_csv(csv_path, header=None)
    return csv_data[column].to_numpy()


def test_ecg_findpeaks_peakdetect_good_4000(benchmark):
    good_4000 = _read_csv_column('good_4000.csv', 1)
    benchmark(_ecg_findpeaks_peakdetect, good_4000)


def test_ecg_findpeaks_peakdetect_bad_500(benchmark):
    bad_500 = _read_csv_column('bad_500.csv', 1)
    benchmark(_ecg_findpeaks_peakdetect, bad_500)
```
</details>

# Checklist


neurokit2/ecg/ecg_findpeaks.py
```diff
@@ -1036,65 +1036,53 @@ def _ecg_findpeaks_MWA(signal, window_size):
...
-                        signal_peaks[-2] = missed_peak
+    last_peak = 0
+    last_index = -1
 
-            else:
-                noise_peaks.append(peak)
-                NPKI = 0.125 * detection[noise_peaks[-1]] + 0.875 * NPKI
+    # TODO: Using plateau_size=(1,1) here avoids detecting flat peaks and
+    # maintains original py-ecg-detectors behaviour, but is not obviously
+    # correct. Should we remove that setting to find also flat peaks?
```
DominiqueMakowski, line 1057: > Should we remove that setting to find also flat peaks?

@JanCBrammer any thoughts?

jukka: I looked at both the Pan-Tompkins and Kalidas-Tamil papers (that describe the algorithms this function is used for) but neither is particularly specific about this detail. Pan-Tompkins just says:

> A peak is a local maximum determined by observing when the
> signal changes direction within a predefined time interval.

In this implementation that "predefined time interval" is just three subsequent samples.

I looked at the data I have, and didn't spot any cases where flat peaks would actually be a problem. It looks like the only situations where flat peak detection would make a difference are things like artefacts where the signal gets saturated at maximum recording amplitude. There's no point in even trying to detect peaks from such data, so I'm not sure if there's any compelling reason to consider changing the `plateau_size` setting. And if we did want to change it, the setting should probably not exceed the length of typical QRS complexes to avoid marking saturation artefacts as peaks.

To summarize, something like `plateau_size=(1, max(1, sampling_rate // 10))` might be reasonable, though not necessarily worth actually implementing.

jukka: I changed the TODO to a NOTE comment in https://github.com/neuropsychology/NeuroKit/pull/450/commits/5a1d82919878e5c781375f18869df8b205eec2c0 based on the above observations.

DominiqueMakowski:
Thanks for digging in and all your explanations :)

> To summarize, something like plateau_size=(1, max(1, sampling_rate // 10)) might be reasonable, though not necessarily worth actually implementing.

What are the downsides of adding it? Also, are the tests expected to fail?

Tam-Pham:
hey @jukka!

Sorry for this late review and thank you for a great PR 😄 

Regarding the flat peaks issue, I did think if it's possible to allow the users to decide if they want flat peaks as peaks or exclude them. However, as you said, ECG peaks are generally quite clear and it's not that often that we come across ECG data with abundant instances of flat peaks. Let's keep this in mind though :) 

Will merge and close this PR 👍

Tam-Pham (approved):

PR #456, opened 2021-05-30
Fix rare peak-inside-padding issues in _ecg_findpeaks_kalidas

jukka:
# Description

Fix rare peak-inside-padding issues in _ecg_findpeaks_kalidas

# Proposed Changes

I occasionally see failures in `kalidas2017` peak detection caused by the algorithm returning peak indices that exceed the length of the input signal.

The cause of this problem is the padding needed by the SWT computation, and the simple fix is to drop the padding after the SWT computation is done.

I can't share the data that triggers this problem, and I couldn't think of an easy way to generate an alternative test case. Thus unfortunately I don't have a test case to go with this change. At least the change itself is very simple, and I've verified it locally with the data I have.

# Checklist

### DIFF
--- a/neurokit2/ecg/ecg_findpeaks.py
+++ b/neurokit2/ecg/ecg_findpeaks.py
@@ -817,7 +817,7 @@ def _ecg_findpeaks_kalidas(signal, sampling_rate=1000):
     b, a = scipy.signal.butter(3, [f1 * 2, f2 * 2], btype="bandpass")
     filtered_squared = scipy.signal.lfilter(b, a, squared)
 
-    # Drop padding to avoid detecting peaks inside it.
+    # Drop padding to avoid detecting peaks inside it (#456)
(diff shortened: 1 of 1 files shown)


neurokit2/ecg/ecg_findpeaks.py
```diff
@@ -815,6 +817,9 @@ def _ecg_findpeaks_kalidas(signal, sampling_rate=1000):
     b, a = scipy.signal.butter(3, [f1 * 2, f2 * 2], btype="bandpass")
     filtered_squared = scipy.signal.lfilter(b, a, squared)
 
+    # Drop padding to avoid detecting peaks inside it.
```
DominiqueMakowski, line 820:
```suggestion
    # Drop padding to avoid detecting peaks inside it (#456)
```

DominiqueMakowski:
Ok I trust your call :) I'll merge once it passes the checks

Issue #465, opened 2021-06-14
bug in _ecg_findpeaks_rodrigues function


FrancisThibaultNRC:
**Describe the bug**
out of bounds happens in the function __ecg_findpeaks_rodrigues while doing something simple.

**To Reproduce**

    duration = 600; sampling_rate = 500; heart_rate = 50; noise = 0.0; method='simple'
    ecg = nk.ecg_simulate(duration=duration,
                          sampling_rate=sampling_rate,
                          noise=noise,
                          heart_rate=heart_rate,
                          method=method,
                          random_state=10)
    ecg_clean = nk.ecg_clean(ecg, sampling_rate=sampling_rate)
    _, neurokit = nk.ecg_peaks(ecg_clean, sampling_rate=sampling_rate, method="rodrigues2021")

**System Specifications**
python 3.8.6
neurokit2 version 0.1.2
OS windows 10

And copy paste the output:
```python
    def test_test():
        """Test if an ecg simulated signal of 50 bpm will result in a mean heart rate feature extraction
         of the same values"""

        plot_simulated_data = False

        # Simulate data: heart beat = 50 and 100 bpm
        duration = 600; sampling_rate = 500; heart_rate = 50; noise = 0.0; method='simple'
        ecg = nk.ecg_simulate(duration=duration,
                              sampling_rate=sampling_rate,
                              noise=noise,
                              heart_rate=heart_rate,
                              method=method,
                              random_state=10)
        ecg_clean = nk.ecg_clean(ecg, sampling_rate=sampling_rate)
>       _, neurokit = nk.ecg_peaks(ecg_clean, sampling_rate=sampling_rate, method="rodrigues2021")
```
```
code\scripts\plugin\test_hrv_plugin.py:1041:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
[...]

PR #478, opened 2021-07-11
Bug fix ecg_findpeaks_rodrigues

TiagoTostas:
Fix bug array out of bounds ecg_findpeaks_rodrigues

# Description
Solve issue #465 by checking the size of the array before stage

### DIFF
--- a/neurokit2/ecg/ecg_findpeaks.py
+++ b/neurokit2/ecg/ecg_findpeaks.py
@@ -935,7 +935,7 @@ def _ecg_findpeaks_WT(signal, sampling_rate=1000):
 # =============================================================================
 
 
-def ecg_findpeaks_rodrigues(signal, sampling_rate=1000):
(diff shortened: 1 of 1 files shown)


neurokit2/ecg/ecg_findpeaks.py
```diff
@@ -935,7 +935,7 @@ def _ecg_findpeaks_WT(signal, sampling_rate=1000):
 # =============================================================================
 
 
-def _ecg_findpeaks_rodrigues(signal, sampling_rate=1000):
+def ecg_findpeaks_rodrigues(signal, sampling_rate=1000):
```
DominiqueMakowski, line 938:
```suggestion
def _ecg_findpeaks_rodrigues(signal, sampling_rate=1000):
```

This needs to remain an internal function since they are called by the main `ecg_findpeaks` :)

Tam-Pham (approved):

PR #499, opened 2021-08-19
More complexity indices

zen-juen:
Add more complexity indices as part of https://github.com/neuropsychology/NeuroKit/projects/2#card-66889896 

- [x] Higuchi's FD
  - added `_fractal_higuchi_optimal_k()` + plot
- [x] Kat's FD 
- [x] Lempel Ziv Complexity


neurokit2/complexity/__init__.py
```diff
@@ -18,6 +18,7 @@
 # Fractal
 from .fractal_dfa import fractal_dfa
 from .fractal_mandelbrot import fractal_mandelbrot
+from .fractal_dimension import fractal_dimension
```
DominiqueMakowski, line 21: I think we should name these directly `fractal_higushi` and `fractal_katz` otherwise it's a bit confusing coz our `fractal_*` already implies "fractal dimension using ... method"

neurokit2/complexity/fractal_dimension.py
```diff
@@ -0,0 +1,122 @@
...
+        dimension = _fractal_dimension_katz(signal)
+
+    return dimension
+        
+
+def _fractal_dimension_higuchi(signal, kmax=5):
+    """Computes Higuchi's Fractal Dimension (HFD) from the time series.
+    Value should fall between 1 and 2.
+    
+    For more information about k parameter selection, see https://ieeexplore.ieee.org/document/7168285
```
DominiqueMakowski, line 55:
```suggestion
    For more information about k parameter selection, see the papers referenced below.
    
    References
    ------------
    - ref + link
```

neurokit2/complexity/fractal_dimension.py
```diff
@@ -0,0 +1,122 @@
...
+            normalization = (N - 1) / (n_max * (k ** 2))
+            Lm_k = np.sum(np.abs(np.diff(signal[m-1::k], n=1))) * normalization
+            sets.append(Lm_k)
+        # Compute average value over k sets of Lm(k)
+        L_k = np.sum(sets) / k
+        average_values.append(L_k)
+
+    # Slope of best-fit line through points
+    k_values = np.arange(1, kmax + 1)
+    slope, _ = - np.polyfit(np.log2(k_values), np.log2(average_values), 1)
```
DominiqueMakowski, line 85: In https://ieeexplore.ieee.org.remotexs.ntu.edu.sg/document/7168285 they say: 

image

should there be a minus 2 somewhere?

neurokit2/complexity/fractal_dimension.py
DominiqueMakowski: That said, we should probably find a way of having some `complexity_k()` optimizer function

neurokit2/complexity/fractal_dimension.py
zen-juen: Haven't looked into this one yet but this may be specifically for the calculation of `k` since other papers documenting the algorithm doesn't mention this :))

PR #504, opened 2021-08-24
0.1.4.1

DominiqueMakowski:


neurokit2/complexity/fractal_dfa.py
```diff
@@ -446,8 +448,11 @@ def _fractal_dfa_fluctuation(segments, trends, multifractal=False, q=2):
         var = np.var(detrended, axis=1)
         # obtain the fluctuation function, which is a function of the windows
         # and of q
+        # ignore division by 0 warning
+        old_setting = np.seterr(divide="ignore", invalid="ignore")
         fluctuation = \
             np.float_power(np.mean(np.float_power(var, q / 2), axis=1), 1 / q.T)
+        np.seterr(**old_setting)
```
DominiqueMakowski, line 455:
```suggestion
        np.seterr(**old_setting)  # restore previous setting
```

neurokit2/complexity/fractal_dfa.py
```diff
@@ -446,8 +448,11 @@ def _fractal_dfa_fluctuation(segments, trends, multifractal=False, q=2):
         var = np.var(detrended, axis=1)
         # obtain the fluctuation function, which is a function of the windows
         # and of q
+        # ignore division by 0 warning
+        old_setting = np.seterr(divide="ignore", invalid="ignore")
```
DominiqueMakowski, line 452: it's quite dangerous to play with settings tho isn't there any other way?

Tam-Pham: i don't really see any warning when I was running with individual data but I saw some warnings when I was running with the batch of data from multiple participants. I cannot pinpoint the problem tho so I thought abt turning off the warning. And I restore the setting immediately after that. Tho if you think we shouldn't fiddle with the settings then I can revert it. It's just a little too many warnings 😅

Tam-Pham: After reviewing the warnings, I noticed that I observed the "division by zero" warning in 
https://github.com/neuropsychology/NeuroKit/blob/abfd08db61d9d9142d796fd2efac149e8f4636e7/neurokit2/complexity/fractal_dfa.py#L368

when the warning occurred in
https://github.com/neuropsychology/NeuroKit/blob/abfd08db61d9d9142d796fd2efac149e8f4636e7/neurokit2/complexity/fractal_dfa.py#L450

so perhaps some instances might return detrended values equal to zero and thus the fluctuation is returned as zero (undefined float_power of zero), leading to an undefined log value (when zero is input to log function, it apparently returns the "division by zero" warning too, as discussed [here](http://5.9.10.113/66844763/numpy-log-on-division-with-0-in-numerator-gives-divide-by-zero-warning)

Tam-Pham:
@DominiqueMakowski do make a review before the patch version is merged

NEWS.rst
```diff
@@ -2,6 +2,17 @@ News
...
 
+
+0.1.4.1
+-------------------
+
+Fixes
++++++++++++++
+* Adjust `kmax` parameter in `fractal_higuchi()` according to signal length as having `kmax` more than half of signal length leads to division by zero error
+* Ensure that sanitization of input in `_hrv_dfa()` is done before windows for `DFA_alpha2` is computed
+* `np.seterr` is added to `fractal_dfa()` to avoid returning division by zero warning which is an unexpected behaviour
```
DominiqueMakowski, line 13:
```suggestion
* `np.seterr` is added to `fractal_dfa()` to avoid returning division by zero warning which is an expected behaviour
```

@Tam-Pham right?

PR #505, opened 2021-08-24
fix fractal_higuchi

zen-juen:
# Checklist

- [ ] I have read the [CONTRIBUTING](https://github.com/neuropsychology/NeuroKit/blob/master/.github/CONTRIBUTING.rst#structure-and-code) file.
- [ ] My PR is targetted at the **dev branch** (and not towards the master branch).

### DIFF
--- a/neurokit2/complexity/fractal_higuchi.py
+++ b/neurokit2/complexity/fractal_higuchi.py
@@ -138,10 +138,8 @@ def _fractal_higuchi_plot(k_values, average_values, kmax, slope, intercept, ax=N
     # Label all values unless len(k_values) > 10 then label only min and max k_max
     if len(k_values) < 10:
         for i in range(0, len(k_values)):
-            ax.scatter(
-                -np.log(k_values[i]), np.log(average_values[i]), color=colors[i],
-                marker='o', zorder=2, label="k = {}".format(i)
-                )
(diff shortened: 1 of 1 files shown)


neurokit2/complexity/fractal_higuchi.py
```diff
@@ -169,24 +169,26 @@ def _fractal_higuchi_optimal_k(signal, k_first=2, k_end=60):
...
-    if optimal_k <= 2:
+    if kmax <= 2:
         warn(
             "The detected kmax value is 2 or less. The change in HFD values across kmax values is dependent "
             " on the fractal source data and there may be no plateau in this case."
             " Consider choosing a manual kmax value.",
             category=NeuroKitWarning
         )
         grad = np.diff(slope_values) / np.diff(k_range)
-        optimal_k = np.where(grad == np.max(grad[2:]))[0][0]
```
Tam-Pham, line 189: I feel like this does not help to address the problem that k_range can still be 2 (since k_first is 2). So I feel like it should work better if u can give the warning and set k_max to a value like 15 or 20 (large enough). 

Then the warning can be:
            "The detected kmax value is 2 or less. The change in HFD values across kmax values is dependent "
            " on the fractal source data and there may be no plateau in this case."
            " HFD value will be calculated with kmax=20. "

PR #520, opened 2021-08-30
0.1.5

DominiqueMakowski:


DominiqueMakowski:
@zen-juen fractal_psdslopes currently requires sampling_rate as an arg but I think we could remove that too without affecting the results right?

zen-juen:
> @zen-juen fractal_psdslopes currently requires sampling_rate as an arg but I think we could remove that too without affecting the results right?

@DominiqueMakowski Nuu this one we need sampling rate because PSD is computed for a specific frequency range and later converted to fractal dimension

Issue #528, opened 2021-09-01
Confusion between exhalation peaks/troughs and inhale/exhale onsets [RSP]


shanelindsay:
Neurokit labels peaks in the respiration signal as inhalation peaks and troughs as exhalation troughs (e.g. the rsp_plot graph). The example https://neurokit2.readthedocs.io/en/latest/examples/rrv.html doesn't specify the nature of the respiration measurment (but perhaps should), though here refers to inspiration peaks and exhalation troughs. 
This had potential for confusion, as peaks and troughs in a respiration signal have different meaning depending on the type of recording and the measure (i.e. volume, spirometer or breathing belt). 
The code which calculates whether a sample is inspiration or expiration assumes that a trough is the start of inhalation, which would the case for breathing belt data (and I assume spirometer users might not be using neurokit for analysis). 

However, some of the code (if I am reading it right) seems to assume that "peaks" are inhalation onsets. For example, 
https://neurokit2.readthedocs.io/en/latest/_modules/neurokit2/rsp/rsp_rrv.html#rsp_rrv
bbi = np.peaks(peaks) / sampling rate * 1000
Here bbi is breath to interval, which is used to calculate a number of measures. 
Traditionally these measures would use the onset of the inhalation as the start of a breath cycle not the start of the exhale (i.e. the "inspiration peak" in neurokit terminology), so it seems the code should be **np.peaks(troughs)**. 
So this suggests a few possible changes - fixing the code to use troughs not peaks (and I think a few other functions are affected), being more explicit about the respiration recording type assumed in the docs, and perhaps shifting more to labelling peaks and troughs as inhale onsets and exhale onsets instead of inspiration/exhalation troughs and inspiration peaks (which would reduce ambiguity of the meaning of a "inspiration/exhalation peak/trough").
[note, some functions might work very similar regardless of using peaks and troughs, but in terms of reporting of measures inhale onsets would be better to fit with standard methods)

zen-juen:
Great call! Interesting that the shading of our rsp plot is actually in line with what was raised here, in that the trough is the start of the breath cycle (and hence inhalation onset), although our operationalization of bbi as `np.diff(peaks)` implies that each peak is the inhalation onset. 
![image](https://user-images.githubusercontent.com/54053748/131770355-6e926ec5-7800-4a87-9fac-8bd6aa386d05.png)

[...]

PR #536, opened 2021-09-05
Add and clarify RSP features

zen-juen:
See https://github.com/neuropsychology/NeuroKit/projects/4#card-67978636 and #528

### DIFF
--- a/neurokit2/rsp/rsp_eventrelated.py
+++ b/neurokit2/rsp/rsp_eventrelated.py
@@ -7,7 +7,7 @@ from ..epochs.eventrelated_utils import (_eventrelated_addinfo,
                                          _eventrelated_rate,
                                          _eventrelated_sanitizeinput,
                                          _eventrelated_sanitizeoutput)
-from ..misc import NeuroKitWarning
+from ..misc import NeuroKitWarning, find_closest
 
 
 def rsp_eventrelated(epochs, silent=False, subepoch_rate=[None, None]):
@@ -40,6 +40,7 @@ def rsp_eventrelated(epochs, silent=False, subepoch_rate=[None, None]):
         - *"RSP_Rate_SD"*: the standard deviation of the respiratory rate after stimulus onset.
         - *"RSP_Rate_Max_Time"*: the time at which maximum respiratory rate occurs.
         - *"RSP_Rate_Min_Time"*: the time at which minimum respiratory rate occurs.
+        - *"RSP_Amplitude_Baseline"*: the respiratory amplitude at stimulus onset.
         - *"RSP_Amplitude_Max"*: the change in maximum respiratory amplitude from before stimulus onset.
         - *"RSP_Amplitude_Min"*: the change in minimum respiratory amplitude from before stimulus onset.
         - *"RSP_Amplitude_Mean"*: the change in mean respiratory amplitude from before stimulus onset.
@@ -122,6 +123,8 @@ def _rsp_eventrelated_amplitude(epoch, output={}):
         return output
 
     # Get baseline
+    zero = find_closest(0, epoch.index.values, return_index=True)  # Find index closest to 0
+
     if np.min(epoch.index.values) <= 0:
         baseline = epoch["RSP_Amplitude"][epoch.index <= 0].values
         signal = epoch["RSP_Amplitude"][epoch.index > 0].values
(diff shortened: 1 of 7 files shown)


zen-juen:
@DominiqueMakowski @shanelindsay I've changed all (I think) instances of "inhalation peaks" and "exhalation troughs", to "peaks (exhalation onsets)" and "troughs (inhalation onsets)" respectively in the docstrings. The explicit labelling of onsets is done only in `rsp_plot` and I left the namings in the generated signal columns/dictionary info as they are (i.e., `RSP_Peaks` and `RSP_Troughs`) since it seems more consistent with our API like that. `rsp_rrv()` also now correctly uses troughs rather than peaks. However, I was wondering does this affect the computation of respiratory rate using `rsp_rate()`? 🤔 There are currently two methods:
>        Method can be either "peak" or "xcorr". In "peak" method, rsp_rate is calculated from the periods between respiration peaks. In "xcorr" method, cross-correlations between the changes in respiration with a bank of sinusoids of different frequencies are caclulated to indentify the principal frequency of oscillation.

The latter is independent of the indices but for the former, does it matter whether periods are computed from peaks or troughs?

neurokit2/rsp/rsp_eventrelated.py
```diff
@@ -130,6 +133,7 @@ def _rsp_eventrelated_amplitude(epoch, output={}):
         signal = epoch["RSP_Amplitude"][epoch.index > np.min(epoch.index)].values
 
     # Max / Min / Mean
+    output["RSP_Amplitude_Baseline"] = epoch["RSP_Amplitude"].iloc[zero]
```
DominiqueMakowski, line 136: since you compute the baseline above why don't you re-use this variable?

zen-juen: They're different, 'baseline' variable is computed for all time points before stimulus onset (t=0) whereas rsp_amplitude_baseline is essentially the value closest to t=0

DominiqueMakowski: riiight. which maeks me think, what's the default epoch window used for ecg_ and rsp_ analysis (and the one that we use in cocon?)

DominiqueMakowski: I wonder whether we should use the "closest to 0" adjustment method for the mean/max/min parameters; because it's based on interpolated signal so the real baseline is the value at 0, averaging some time before makes little sense (the pre-0 window can be used for stuff like detrending or whatnot, but for baseline correction I'd say we just correct with the closest to 0 value)

zen-juen: > riiight. which maeks me think, what's the default epoch window used for ecg_ and rsp_ analysis (and the one that we use in cocon?)

There were two window lengths we used from what we discussed last time, so shorter window is -0.5 to 1.5 for both ECG and RSP, and longer window is -0.5 to 2.25s for ECG and -0.5 to 3s for RSP.

> I wonder whether we should use the "closest to 0" adjustment method for the mean/max/min parameters; because it's based on interpolated signal so the real baseline is the value at 0, averaging some time before makes little sense (the pre-0 window can be used for stuff like detrending or whatnot, but for baseline correction I'd say we just correct with the closest to 0 value)

Yes I agree, for our rate features the mean/max/min parameters are corrected based on this exact 'closest to 0' value, wonder why we didn't do this for our amplitude features 🤔

zen-juen: @DominiqueMakowski bump ^^

DominiqueMakowski: yeah we should be consistent and hte the closest to 0 point

zen-juen:
> However, I was wondering does this affect the computation of respiratory rate using `rsp_rate()`? 🤔 There are currently two methods:

To follow up on the question I raised earlier, here's a visualization of the differences in rate computed from peaks vs. troughs
Figure_1

DominiqueMakowski:
orange is peaks or troughs?

zen-juen:
Oops I can't remember but here's a new short example 😄 
```
rsp = nk.rsp_simulate(duration=100)
rsp_cleaned = nk.rsp_clean(rsp, sampling_rate=1000)
_, info = nk.rsp_peaks(rsp_cleaned, sampling_rate=1000)
rate_from_peaks = nk.signal_rate(info["RSP_Peaks"], sampling_rate=sampling_rate, desired_length=len(rsp_cleaned))
rate_from_troughs = nk.signal_rate(info["RSP_Troughs"], sampling_rate=sampling_rate, desired_length=len(rsp_cleaned))
nk.signal_plot([rate_from_peaks, rate_from_troughs], labels=['Peaks', 'Troughs'])
```

Figure_2

DominiqueMakowski:
I'd go with throughs, since it's true that a "breathing cycle" would start with inspiration (and hence a through), so it would conceptually make sense to have rsp rate defined as trough to trough. But then this is from the top of my head, I'm sure in papers they talk about how they compute it (in the case of respi belts)

zen-juen:
See https://ieeexplore.ieee.org/abstract/document/7985893 🤔 (and https://www.degruyter.com/document/doi/10.1515/cdbme-2016-0054/html)
>Researchers have tried different approaches to estimate the
respiration rate from the extracted respiration signal. Based on
the literature, the use of the signal auto-correlation has proven
to be efficient for respiration rate estimation [13]–[15]. In this
paper, the auto-correlation is used to estimate the respiration
rate from the signals produced by both combining methods, as
well as for the gold standard signal from the belt system

PR #550, opened 2021-09-15
Compute respiration rate based on troughs

zen-juen:
Have respiratory rate computed based on troughs instead of peaks as discussed in #528

### DIFF
--- a/neurokit2/rsp/rsp_rate.py
+++ b/neurokit2/rsp/rsp_rate.py
@@ -6,15 +6,18 @@ from ..signal import (signal_resample, signal_rate, signal_interpolate,
 from .rsp_peaks import rsp_peaks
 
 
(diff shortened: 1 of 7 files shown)


neurokit2/rsp/rsp_rate.py
```diff
@@ -6,30 +6,34 @@
...
-        the principal frequency of oscillation.
-    peak_method : str
-        Method to identify respiration peaks. Can be one of "khodadad2018" (default) or "biosppy".
+        Method can either be "period" or "xcorr". In "period" method, respiratory rate is calculated from the
+        periods between successive inspirations (i.e., inhalation onsets/troughs).
+        In "xcorr" method, cross-correlations between the changes in respiration with a bank of
+        sinusoids of different frequencies are caclulated to indentify the principal frequency of oscillation.
+    period_method : str
+        Method to identify successive respiratory inspirations, only relevant if method is "period".
+        Can be one of "khodadad2018" (default) or "biosppy".
```
zen-juen, line 34: Change naming of method from "peaks" to "periods" so that it's clearer

PR #551, opened 2021-09-16
Functions to account for more `mne` features

zen-juen:
- [x] Account for `first_samp` attribute in mne channels
`mne` raw objects have an attribute called `first_samp` corresponding to the time lapsed after the system start and before the start of the recording, and so the first event's sample number is always > 0. This is taken into account for in mne functions but not in neurokit yet 

See http://predictablynoisy.com/mne-python/generated/mne.io.RawArray.html

- [x] `epochs_plot()` to work with mne objects

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -16,6 +16,11 @@ New Features
 
 * Updates logic to find `kmax` in `fractal_higuchi()`
 * Add RSP_Amplitude_Baseline in event-related analysis
+* Add argument `add_firstsamples` in `mne_channel_extract()` to account for first sample attribute in mne raw objects
+* Allow plotting of `mne.Epochs` in `epochs_plot()`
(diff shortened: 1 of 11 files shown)


DominiqueMakowski:
@zen-juen could you add a `show=False` argument to eeg_badchannels taht would plot all the channels overlayed together in like thin dark grey and the channels detected as "bad" as red on top of them

zen-juen:
> @zen-juen could you add a `show=False` argument to eeg_badchannels taht would plot all the channels overlayed together in like thin dark grey and the channels detected as "bad" as red on top of them

Figure_1

Currently looks like this

DominiqueMakowski:
Noice

zen-juen:
If several bad channels

Figure_1

PR #567, opened 2021-10-30
Complexity hotfixes

DominiqueMakowski:


DominiqueMakowski:
```python
import neurokit2 as nk

signal = nk.signal_simulate(duration=2, sampling_rate=200, frequency=[5, 6], noise=0.5)

mplzc, info = nk.complexity_lempelziv(signal, delay=7, dimension=3, multiscale=True, show=True)
```

MPLZC fails as soon as delay get more than like 1 or 2 🤔

zen-juen:
> ```python
> import neurokit2 as nk
> 
> signal = nk.signal_simulate(duration=2, sampling_rate=200, frequency=[5, 6], noise=0.5)
> 
> mplzc, info = nk.complexity_lempelziv(signal, delay=7, dimension=3, multiscale=True, show=True)
> ```
> 
> MPLZC fails as soon as delay get more than like 1 or 2 🤔

That's because for the final scale factor, the length of the coarse-grained signal (which depends on dimension and the scale factors we use) cannot be shorter than 2 times the delay (so for the above example the final coarsegrained signal has a length of 12 data samples and embedding delay of lower than 12 / 2 = 6 is needed and hence `delay=7` raises the embedding error).

On a related note, I found the reviewers report to the original MPLZC paper [here](https://www.mdpi.com/1099-4300/23/7/832/review_report). This is the author's response to the reviewers' comments regarding how the role of tau is unclear in the paper:
> The unity delay (τ=1) is used to calculate the PLZC of the coarse-grained time series for all scales. There are several methods for determining the optimal time delay, for a single channel. Future developments should test time delay selection for different scales, but these are not yet available.

DominiqueMakowski:
should we use always delay=1 when coarsegraining?

zen-juen:
Yes perhaps we should emphasize in our docstrings to use `delay=1` when `multiscale=True` for now

DominiqueMakowski:
for the other multiscale measures it just hardcodes delay=1 when doing the coarsegraining (overwriting in the process any other value) we should probably just do the same right?

zen-juen:
ah yes you're right let's hardcode it then!

docs/readme/README_examples.py
```diff
@@ -5,6 +5,7 @@
 import matplotlib.pyplot as plt
 import matplotlib.cm
 import neurokit2 as nk
+import PyEMD
```
DominiqueMakowski, line 8: what's dis

zen-juen: `signal_decompose` in readme need dis

zen-juen: wait how come last time readme works without importing this

PR #569, opened 2021-11-04
Fix matrix B created too short

pjercic:
# Description

This PR aims at adding this fix where the data matrix B for creating the sparse matrix D_2 was created too short (N-2) and it didn't reach the two last columns in the sparse matrix upon creation.

# Proposed Changes

I changed the `_signal_detrend_tarvainen2002(signal, regularization=500)` function so that I extended the matrix B to fix the mentioned issue.

# Checklist

### DIFF
--- a/neurokit2/signal/signal_detrend.py
+++ b/neurokit2/signal/signal_detrend.py
@@ -134,7 +134,7 @@ def _signal_detrend_tarvainen2002(signal, regularization=500):
     """
     N = len(signal)
     identity = np.eye(N)
-    B = np.dot(np.ones((N - 2, 1)), np.array([[1, -2, 1]]))
+    B = np.dot(np.ones((N, 1)), np.array([[1, -2, 1]]))
     D_2 = scipy.sparse.dia_matrix((B.T, [0, 1, 2]), shape=(N - 2, N))  # pylint: disable=E1101
     inv = np.linalg.inv(identity + regularization ** 2 * D_2.T @ D_2)
     z_stat = ((identity - inv)) @ signal


DominiqueMakowski:
Hi @pjercic thanks for this PR. 

I am not sure about that fix since this is an existing algorithm which is implemented in a similar way elsewhere; e.g.:

- https://github.com/rhenanbartels/hrv/blob/190c923250884b1f5632e38658bd8df7d9d5352e/hrv/detrend.py#L115
- https://github.com/LucaCerina/HRV-detrend/blob/master/detrendFast.m#L32

I don't remember the original paper, but if you could check that source that'd be great, so that we know whether it is the implementations that are possibly wrong, or perhaps the original algo is bug-prone too.

In what context did you obtain an error?

JanCBrammer:
The algorithm seems to be published here: https://sci-hub.st/10.1109/10.979357

Tarvainen, M. P., Ranta-aho, P. O., & Karjalainen, P. A. (2002). An advanced detrending method with application to HRV analysis. IEEE Transactions on Biomedical Engineering, 49(2), 172–175. doi:10.1109/10.979357

DominiqueMakowski:
Thanks @JanCBrammer, so we shouldn't touch it in your opinion?

JanCBrammer:
I think that depends on whether or not the original reference is correct. @pjercic, the bug might as well originate there?

pjercic:
I hit an error using your code to do smooth priors de-trending, and pinpointed it to the issue of the generator data for creating matrix being too short.
Therefore I fixed it and it also fixed the output problem specified here.

I am not sure where the error ended up in your pipeline with calling library errors, but the real failed case is of course, and assert values, since the output of this specific de-trend was not correct.
FAILED tests/tests_signal.py::test_signal_detrend - assert False

DominiqueMakowski:
Pinging @rhenanbartels for his opinion about that `-2` since we currently share the same implementation :)

pjercic:
There have been further developments where another person suggested a further improvement.
I need to test on my side, but it is a sound proposal, check it out.

https://stackoverflow.com/questions/69810723/issues-with-signal-detending-using-smoothness-priors-on-the-last-detended-elemen/74365247#74365247

neurokit2/signal/signal_detrend.py
```diff
@@ -134,7 +134,7 @@ def _signal_detrend_tarvainen2002(signal, regularization=500):
     """
     N = len(signal)
     identity = np.eye(N)
-    B = np.dot(np.ones((N - 2, 1)), np.array([[1, -2, 1]]))
+    B = np.dot(np.ones((N, 1)), np.array([[1, -2, 1]]))
     D_2 = scipy.sparse.dia_matrix((B.T, [0, 1, 2]), shape=(N - 2, N))  # pylint: disable=E1101
```
DominiqueMakowski, line 138: Interesting, so they suggest changing that line to:

```suggestion
    D_2 = scipy.sparse.dia_matrix((B.T, [0, 1, 2]), shape=(N - 3, N - 1))
```

DominiqueMakowski: Please do give it a test. Does this change a lot in "normal" scenarios (i.e., would user that used the previous version have significantly different results after the change)? Just to know wether to label that change as a fix or a breaking change

PR #571, opened 2021-11-04
Noise simulation

DominiqueMakowski:


DominiqueMakowski:
> Nuu this one we need sampling rate because PSD is computed for a specific frequency range and later converted to fractal dimension

@zen-juen I'm pretty sure that if we hardcode the SR the results don't change though ^^ could you try pliz and confirm or INFIRM my hypothesis

zen-juen:
@DominiqueMakowski yes yes it doesn't change the actual dimension value (since it's obtained from plotting log power vs freq) but it does alter the frequency axis here in our plot ^^
```
nk.fractal_psdslope(signal, sampling_rate=200, frequency_range='lowest25')
nk.fractal_psdslope(signal, sampling_rate=50, frequency_range='lowest25')
```
image

DominiqueMakowski:
yeah but these values are pretty meaningless anyway, so it's okay (because we didn't specify the unit anyway in the x-axis title so it's still correct). So we can just remove the sampling rate 😁 👍

zen-juen:
@DominiqueMakowski 🙄

DominiqueMakowski:
- [ ] nld with short signals gives an unclear warning, not sure wether we want to make it more explicit like psdslopes or what:

image

- [ ] @Tam-Pham hrv_time line 159 and 175, in the `_sdann()` we have sampling_rate as an *unused* arg, but shouldn't it be used (where currently it multiplies by 1000)
- [ ] @zen-juen a (new?) bug in dimension optim:

image

zen-juen:
@DominiqueMakowski nono because before this line we have `np.array_split()` which makes this length 10 signal into for lengths 4, 3, 3 so it includes everything liao

```
np.array_split(x, n_epochs)
[array([0.        , 0.23272573, 0.44840112, 0.63150375]),
 array([0.76942088, 0.85355339, 0.88003676]),
 array([0.85001176, 0.76942088, 0.64835267])]
```

DominiqueMakowski:
Interesting, so the first epoch is a big longer? then it's okay, since it's consistent and only marginally bigger

zen-juen:
> Interesting, so it's always the first epoch that is eventually a big longer? then it's okay, since it's consistent and only marginally bigger

korekkorek 😄

zen-juen:
As for this specific bug in `complexity_optimize()` (one of the internal functions)

```
    d, dist, index, y2 = _embedding_dimension_d(signal, dimension, delay, metric, window, maxnum)

    # Compute the ratio of near-neighbor distances in d + 1 over d dimension
    # Its average is E(d)
    E = np.mean(d / dist)

    # Calculate E^*(d)
    Es = np.mean(np.abs(y2[:, -1] - y2[index, -1]))

```
with `dist` = distance of nearest neighbours in dim 1 and `d` = distance of nearest neighbours in dim 2: it seems like in both arrays there are some distances = 0 hence the divide by zero error. this then causes E1 to contain nan:

```
        min_dimension = [i for i, x in enumerate(E1 >= 0.85 * np.max(E1)) if x][0] + 1
```
so this will error, though one quick way around it is just to have `np.nanmax(E1)` here? 🤔 I'm not sure how else we should go about fixing this because removing distances of 0 doesn't really seem right as well

DominiqueMakowski:
what if we set E to np.nan? would it mess it up downstream?

zen-juen:
> what if we set E to np.nan? would it mess it up downstream?

What do you mean? Like if there is any d = 0 or dist = 0 then E = np.nan if not E = np.mean(d / dist)?

DominiqueMakowski:
yea

zen-juen:
So with your simulated signal example, it'll return

```
nk.complexity_dimension(signal, delay=delay, show=True)
Out[307]: 
(6,
 {'Method': 'afnn',
  'Values': array([ 1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11, 12, 13, 14, 15, 16, 17,
         18, 19, 20]),
  'E1': array([       nan, 0.1466981 , 0.6500627 , 0.50930644, 0.76639426,
         0.87172263, 0.91888627, 0.9638453 , 0.91463178, 0.97797958,
         0.97195816, 0.99784511, 1.00318121, 0.96001641, 0.97681836,
         0.9946773 , 0.99486582, 0.9941719 , 0.98702221, 0.99896367]),
  'E2': array([9.73101546e-14, 1.15066406e+00, 1.02878675e+00, 8.61136963e-01,
         9.39325502e-01, 1.01145485e+00, 1.03146911e+00, 9.99814646e-01,
         9.79044168e-01, 1.01501939e+00, 1.03365291e+00, 1.01418238e+00,
         1.01934230e+00, 1.00496707e+00, 1.00689774e+00, 1.03147898e+00,
         1.01377595e+00, 1.00661703e+00, 1.01899047e+00, 1.01889308e+00])})
```

DominiqueMakowski:
I guess it's good as it doesnt seem to break the optimal dimension detection

Issue #560, opened 2021-10-19
signal_psd() pesky warning


DominiqueMakowski:
When using `signal_psd()` with Welch method I pretty much always get the following warning, which is a bit obscure. We should somehow improve that

`The duration of recording is too short to support a sufficiently long window for high frequency resolution. Consider using a longer recording or increasing the min_frequency`

DominiqueMakowski:
Somewhat related to https://github.com/neuropsychology/NeuroKit/issues/466

DominiqueMakowski:
On top of that, I note that base R gives way nicer results that python:

``` r
signal = c(3, 3, 5, 7, 1, 3, 5, 3, 5, 6, 1, 3, 5, 3)
y <- stats::ts(signal, frequency = 100)
npad <- stats::nextn(length(signal), factors = 2)
psd  <- stats::spec.pgram(y, fast = FALSE, plot=FALSE, pad=npad, taper=0.5)
plot(psd$freq, psd$spec)
```

![](https://i.imgur.com/JVDfndS.png)

<sup>Created on 2021-10-23 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


```python
signal = np.array([3, 3, 5, 7, 1, 3, 5, 3, 5, 6, 1, 3, 5, 3])
sampling_rate = 100
psd = nk.signal_psd(signal, sampling_rate=sampling_rate, method = "fft")
plt.plot(psd["Frequency"], psd["Power"])
```

![image](https://user-images.githubusercontent.com/8875533/138535185-801e5b48-e8d9-41f3-a5ed-cc7f86810795.png)


I suspect it might have to do with the ` stats::nextn(length(signal), factors = 2)` setting for the npadding but I can't find any equivalence in Python...

PR #574, opened 2021-11-06
No warning for signal_psd: default min_freq

Tam-Pham:
warning for low-frequency resolution
Address #560 

# Checklist


DominiqueMakowski:
But so you agree that currently, by default, in most of the cases: `nperseg = int(len(signal) / 2)` right?

DominiqueMakowski:
why not "sanitize" min_freq to the lowest possible instead of 0.001

DominiqueMakowski:
doing: `min_freq = max(min_freq, min_possible_freq_given_length_and_SR)` and then padding from 0 ?

DominiqueMakowski:
We can sanitize min_freq like so, assuming it is 0 by default:

`min_freq = max(min_freq, 1/(len(signal) / sampling_rate))`

Tam-Pham:
> But so you agree that currently, by default, in most of the cases: nperseg = int(len(signal) / 2) right?

No 😅 only when the recording length is not sufficiently long to capture at least 2 cycles of min_frequency

I don't know if you are misunderstanding but there is no "impossible frequencies". We are just using the min_frequency to determine the length of overlapping segments when calculating PSD.

DominiqueMakowski:
>  only when the recording length is not sufficiently long

How come i get the warning *everytime* I use signal_psd then??

Tam-Pham:
> How come i get the warning _everytime_ I use signal_psd then?? even for signals of length a million

Because if min_frequency is zero (which is adjusted to 0.01), the signal length required for "absolute" high level of good resolution is 4 million 😅 

For your use case, you don't have a default value of min_frequency that you want?

DominiqueMakowski:
hence the current "adjusted" value is useless since it requires a very rare usecase

All I want is a default behaviour that 1) doesnt' require to manually input min_freq and 2) doesn't print a warning for 99.999% of cases 😅

Tam-Pham:
> All I want is a default behaviour that 1) doesnt' require to manually input min_freq and 2) doesn't print a warning for 99.999% of cases 😅

What we can do is add a "default" option for min_frequency where min_frequency is determined by optimal nperseg (based on sampling_rate and length of the signal.

DominiqueMakowski:
> What we can do is add a "default" option

neurokit2/complexity/entropy_multiscale.py
```diff
@@ -250,6 +244,15 @@ def _entropy_multiscale_cmse(signal, tau, dimension, tolerance, fuzzy, **kwargs)
             y[i, :], delay=1, dimension=dimension, tolerance=tolerance, fuzzy=fuzzy, **kwargs
         )[0]
 
+    if len(np.where((mse_y == np.inf) | (mse_y == -np.inf) | (mse_y == np.nan))[0]) == len(mse_y):
+        # return nan if all are infinity/nan values
+        return np.nan
+    else:
+        # Remove inf, nan and 0
+        mse_y = mse_y[~np.isnan(mse_y)]
```
DominiqueMakowski, line 252: can't do one liner? `mse_y[(~np.isnan(mse_y)) | (mse_y != np.inf) | (mse_y != -np.inf)]`

zen-juen: u right ^^

DominiqueMakowski:
@Tam-Pham currently we "normalize" the power in many different places (in signal_psd), can we unify that so that it's done only at one place?

DominiqueMakowski:
Why this so nice:

```python
import neurokit2 as nk
import mne
raw.plot_psd(fmin=0, fmax=40., picks=["EEG 050"])
```

image


And ours so ugly 🥲 

```python
channel = nk.mne_channel_extract(raw, what=["EEG 050"]).values
psd = nk.signal_psd(
    channel, sampling_rate=raw.info["sfreq"], show=True, max_frequency=40, method="multitapers"
)
```

image

DominiqueMakowski:
> And ours so ugly 🥲

what to do lah

Tam-Pham:
Next PR? 🥲

PR #592, opened 2021-12-07
Support for reading multiple devices + events in `read_bitalino`

zen-juen:
- [ ] I have read the [CONTRIBUTING](https://github.com/neuropsychology/NeuroKit/blob/master/.github/CONTRIBUTING.rst#structure-and-code) file.
- [ ] My PR is targetted at the **dev branch** (and not towards the master branch).

### DIFF
--- a/neurokit2/complexity/complexity_embedding.py
+++ b/neurokit2/complexity/complexity_embedding.py
@@ -33,7 +33,9 @@ def complexity_embedding(signal, delay=1, dimension=3, show=False, **kwargs):
     Parameters
     ----------
     signal : Union[list, np.array, pd.Series]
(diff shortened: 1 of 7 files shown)


zen-juen:
A rather lengthy change to `read_bitalino` due to the possibility of having multiple devices (all of which each device may have different sensors / channel names / event annotations etc). Now the function returns two things:

(1) the signals - which takes the form of a pandas df if there was only one device, and a dictionary of pandas df with each dict key corresponding to the device name if multiple devices were read; if `events_annotation` is set to True then columns of photosensor-like signals are returned too, column names being e.g. "A1_label1_start", "A1_label1_end", "A2_label1_start" etc (*channelname_annotatedkey_* because there may be different annotations for different channels) Rather verbose I must admit :(

(2) info dump -  metadata containing the sensors ("ECGBIT, "RESPBIT" etc), corresponding channel names ("A1", "A2" etc), sampling rate, and the events annotation timings  if `events_annotation` is True (I thought it might be a good idea to return all these in case different sensors were attached to different channels across multiple devices)

Disclaimer: not *super* urgent for now 😄

DominiqueMakowski:
can we merge that?

PR #607, opened 2022-01-17
[Change] promac parameters exposition / doc update

levsilva:
# Description

This PR aims at exposing the parameters of "promac" peak detector.

# Proposed Changes
- Three parameters were exposed:
  + peak detectors included in the promac pipeline
  + tolerance threshold
  + SD of Gaussian used for convolution

- **kwargs had to be included in the top-level "ecg_peaks" function
- Names of all peak detectors are now controlled by a global dict
- Two auxiliary functions for "promac" were merged for more consistency
- docstring for "promac" was created


levsilva:
The documentation of "ecg_peaks" and "ecg_findpeaks" are essentially the same. I was thinking in merging those two functions. What do you think about the following plan?

1) Move all functions from "ecg_findpeaks.py" to "ecg_peaks.py" and exclude the first one.
2) Merge functions "ecg_findpeaks" and "ecg_peaks" into "ecg_peaks", since it is on the top.
3) Update usages of "ecg_findpeaks" to "ecg_peaks" throughout neurokit.

DominiqueMakowski:
Awesome, I'll start reviewing asap! 

Regarding the merge of findpeaks and peaks, I am not sure it's a good idea. Basically it used to be the same function, but then ecg_peaks also adds on peak correction which is a heavy process and in some cases we only use the raw peak detection, hence the two different functions. Essentially, users are meant to use _peaks, but for developers, it can be convenient to use _findpeaks in some cases. That said, the documentation should indeed be improved to reflect that!

neurokit2/ecg/ecg_findpeaks.py
```diff
@@ -8,7 +8,7 @@
 from ..signal import signal_findpeaks, signal_plot, signal_sanitize, signal_smooth, signal_zerocrossings
 
 
-def ecg_findpeaks(ecg_cleaned, sampling_rate=1000, method="neurokit", show=False):
+def ecg_findpeaks(ecg_cleaned, sampling_rate=1000, method="neurokit", show=False, **kwargs):
```
DominiqueMakowski, line 11: kwargs should be mentioned in the docstrings

neurokit2/ecg/ecg_findpeaks.py
```diff
@@ -131,34 +131,13 @@ def ecg_findpeaks(ecg_cleaned, sampling_rate=1000, method="neurokit", show=False
...
-    elif method in ["kalidas2017", "swt", "kalidas", "kalidastamil", "kalidastamil2017"]:
-        rpeaks = _ecg_findpeaks_kalidas(ecg_cleaned, sampling_rate)
-    elif method in ["martinez2003", "martinez"]:
-        rpeaks = _ecg_findpeaks_WT(ecg_cleaned, sampling_rate)
-    elif method in ["rodrigues2020", "rodrigues2021", "rodrigues", "asi"]:
-        rpeaks = _ecg_findpeaks_rodrigues(ecg_cleaned, sampling_rate)
-    elif method in ["promac", "all"]:
-        rpeaks = _ecg_findpeaks_promac(ecg_cleaned, sampling_rate=sampling_rate, threshold=0.33, show=show)
+    if method in all_methods.keys():
+        rpeaks = all_methods[method](ecg_cleaned, sampling_rate=sampling_rate, show=show, **kwargs)
```
DominiqueMakowski, line 135: That's a good idea but I wonder if it makes it less verbose in the end since there is the big dictionary that takes even more space 🤔 are there any other advantages to this change?

neurokit2/ecg/ecg_findpeaks.py
```diff
@@ -169,18 +148,66 @@ def ecg_findpeaks(ecg_cleaned, sampling_rate=1000, method="neurokit", show=False
...
-    x = _ecg_findpeaks_promac_addmethod(signal, sampling_rate, x, _ecg_findpeaks_ssf, **kwargs)
-    x = _ecg_findpeaks_promac_addmethod(signal, sampling_rate, x, _ecg_findpeaks_engzee, **kwargs)
-    x = _ecg_findpeaks_promac_addmethod(signal, sampling_rate, x, _ecg_findpeaks_elgendi, **kwargs)
-    x = _ecg_findpeaks_promac_addmethod(signal, sampling_rate, x, _ecg_findpeaks_kalidas, **kwargs)
-    x = _ecg_findpeaks_promac_addmethod(signal, sampling_rate, x, _ecg_findpeaks_WT, **kwargs)
-    x = _ecg_findpeaks_promac_addmethod(signal, sampling_rate, x, _ecg_findpeaks_rodrigues, **kwargs)
+    for method in promac_methods:
+        try:
+            x = _ecg_findpeaks_promac_addconvolve(
+                signal, sampling_rate, x, all_methods[method], gaussian_sd=gaussian_sd, **kwargs
```
DominiqueMakowski, line 205: I see, the global dict is re-used here

neurokit2/ecg/ecg_findpeaks.py
```diff
@@ -1095,3 +1126,45 @@ def _ecg_findpeaks_peakdetect(detection, sampling_rate=1000):
             NPKI = 0.125 * peak_value + 0.875 * NPKI
 
     return signal_peaks
+
+
+# Global dict mapping the method's name to its function
+all_methods = {
```
DominiqueMakowski, line 1132: I wonder if instead of a global dict, it wouldn't be easier and a bit shorter to have an internal function like `_ecg_findpeaks_findmethod(method)` that would do the if-elsing (which seems a bit safer than calling the output of a as one can in the future add some more steps within the block), and that we could re-call in promac. What do you think?

levsilva: This is a good idea. I will change it in a new commit.

levsilva:
I could not find the reference for "martinez03" peak detector. Would it be the same paper for the ECG Delineation algorithm?

>J. P. Martinez, R. Almeida, S. Olmos, A. P. Rocha and P. Laguna, "A wavelet-based ECG delineator: evaluation on standard databases," in IEEE Transactions on Biomedical Engineering, vol. 51, no. 4, pp. 570-581, April 2004, doi: 10.1109/TBME.2003.821031.

DominiqueMakowski:
Hi @levsilva, apologies for the long silence it was a bumpy start of the year for me. 

> Would it be the same paper for the ECG Delineation algorithm?

I don't know, but i'd say it's very likely so we can add that in

I'll continue reviewing and will merge asap

DominiqueMakowski (commented):
looks good :) Thanks

levsilva:
Great! Thanks @DominiqueMakowski.

PR #612, opened 2022-01-20
Replace ba output with sos for ECG filters and determine initial conditions

### LINKED ISSUE #609: Filters do not compute the initial conditions
I started using neurokit for a paradigm exploiting the ECG signal in real-time. I want to use `'hamilton2002'` method. As I am working on a real-time paradigm, data is coming by chunks in a buffer initialized to 0. Thus, when filtering, the filter response introduces a massive artifact at the beginning of all my windows because of the DC component. 

(On the left, one of the first buffers where you still see the DC step; On the right, a buffer a few seconds later)

image

https://github.com/neuropsychology/NeuroKit/blob/d01111b9b82364d28da01c002e6cbfc45d9493d9/neurokit2/ecg/ecg_clean.py#L200-L209

I see that you are using everywhere `ba` output and that you are not computing the initial conditions. I propose to replace the output `ba` with `sos` that is more numerically stable and recommended and to compute the initial conditions with [`sosfilt_zi`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.sosfilt_zi.html?highlight=filt_zi#scipy.signal.sosfilt_zi).

mscheltienne:
Closes #609

I replaced `ba` output with `sos` when needed, and I added the initial condition `zi` if `filt` was used instead of `filtfilt`.
Even if you want to replicate as closely as possible a method, an implementation, IMO, using `sos` instead of `ba` should always be preferred as it should not induce significant differences while being more numerically stable. 

I didn't see I had to branch from `dev` and not `master`; let's see if any conflicts arise.
Also, I added `.DS_Store` to the .gitignore for macOS developers.


mscheltienne:
The only test is the docs? There are no CIs?
I have to say I did not look into the existing tests.

mscheltienne:
Also removing the `[data]` keyword for mne optional dependency; it's now empty and dependency for datasets are installed by default.

https://github.com/mne-tools/mne-python/blob/89e163f60be122be2150ee93a9d3fcea5641b8b7/setup.py#L63-L69

mscheltienne:
Sorry, my bad; reverted that last change since it's obviously not effective until MNE v1.0..

mscheltienne:
I had a quick look at the error, probably `sos` or `zi_coeff` that has to be reshaped for the `zi` if `data` is not a 1D array.
I'll fix it in the beginning of February.

mscheltienne:
I've been chasing after that one for a while, and I don't get it. 
Please have a look, maybe with some prior knowledge on what the code should do it will be easier to fix.

`test_ecg.py` -> `test_ecg_process()`

```
import neurokit2 as nk

sampling_rate = 1000
noise = 0.05

ecg = nk.ecg_simulate(sampling_rate=sampling_rate, noise=noise)
signals, info = nk.ecg_process(ecg, sampling_rate=sampling_rate, 
                               method="neurokit")
```

raises: `TypeError: unsupported operand type(s) for /: 'str' and 'int'`.

One problem is in `ecg/ecg_quality.py`, in `_ecg_quality_averageQRS`, the columns in data have a `dtype` of `string`.

```
data.dtypes
Time
-0.299358    string
-0.298356    string
-0.297355    string
-0.296354    string
-0.295352    string
 
 0.423648    string
 0.424650    string
 0.425651    string
 0.426653    string
 0.427654    string
Length: 727, dtype: object
```

That error looks unrelated to the changes in this PR.. and a pain to understand/fix.

DominiqueMakowski:
Hey @mscheltienne thanks again for your suggestions and help! I've finally found time to look into that and fixed the issue on dev, so I should be able to merge your PR as soon as it passes the checks :)

mscheltienne:
Thanks, personally I would use an assert with a 'sanity-check' message instead of printing 'wtf'; but I guess it does the same thing 😆

mscheltienne:
@DominiqueMakowski one dumb mistake 😅 good catch ;)

DominiqueMakowski:
I'll merge now, hopefully everything works out :) thanks again!

PR #627, opened 2022-03-26
[Feature] add min, max, 80th & 20th percentiles as HRV features

danibene:
Add the minimum, maximum, 20th percentile, and 80th percentile NN as features in the hrv_time function.

- [x] I have read the [CONTRIBUTING](https://github.com/neuropsychology/NeuroKit/blob/master/.github/CONTRIBUTING.rst#structure-and-code) file.
- [x] My PR is targeted at the **dev branch** (and not towards the master branch).
- [x] I ran the [CODE CHECKS](https://github.com/neuropsychology/NeuroKit/blob/master/.github/CONTRIBUTING.rst#run-code-checks) on the files I added or modified and fixed the errors.
- [x] I have added the newly added features to **News.rst** (if applicable)

PR #647, opened 2022-05-27
[Feature] multiple methods of outlier identification

danibene:
# Description
Allow for different methods of outlier identification based on standardization or percentiles.

# Proposed Changes
- Add new outlier identification method based on percentiles: `_find_outliers_percentile()`, defaults to identifying observations below Q1 - 1.5 * IQR or above Q3 + 1.5 * IQR as outliers, e.g. as in [this paper](https://pubmed.ncbi.nlm.nih.gov/30833637/)
- Add argument `method` in `find_outliers()` calling `_find_outliers_standardize()` or `_find_outliers_percentile()`
- Pass keyword arguments to `standardize()` e.g. robust=True

# Checklist

  - Just in case someone else runs into this: I had [an import error with black](https://github.com/psf/black/issues/3088) at first, but it’s all good now
  - Also I decided to ignore the following pylint output since it concerned a line that I did not add (`from ..stats import standardize`)
    ```
    ************* Module find_outliers
    find_outliers.py:4:0: E0402: Attempted relative import beyond top-level package (relative-beyond-top-level)
    ```
  - Though I am confused about which version the changes should correspond to: for the [previous PR](https://github.com/neuropsychology/NeuroKit/pull/627), I listed the features under 0.1.6 in the [News file](https://github.com/neuropsychology/NeuroKit/blob/master/NEWS.rst), but I saw that they were part of [release 0.2.0](https://github.com/neuropsychology/NeuroKit/releases). Does that mean I should move the previous changes to 0.2.0 and this PR to 0.2.1?

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -55,6 +55,7 @@ New Features
 * Plotting function in `eeg_badchannels()` to visualize overlay of individual EEG channels and highlighting of bad ones
 * Add `eog_peaks()` as wrapper for `eog_findpeaks()`
 * Allow `ecg_delineate()` to account for different heart rate
+* Add argument `method` in `find_outliers()` calling `_find_outliers_standardize()' or `_find_outliers_percentile()'
 
 
 Fixes
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -15,6 +15,7 @@ New Features
 +++++++++++++
 
 * Add new time-domain measures in `hrv_time()`: `Prc20NN`, `Prc80NN`, `MinNN`, and `MaxNN`
+* Add argument `method` in `find_outliers()` calling `_find_outliers_standardize()` or `_find_outliers_percentile()`
 
 Fixes
 +++++++++++++
@@ -55,7 +56,6 @@ New Features
 * Plotting function in `eeg_badchannels()` to visualize overlay of individual EEG channels and highlighting of bad ones
 * Add `eog_peaks()` as wrapper for `eog_findpeaks()`
 * Allow `ecg_delineate()` to account for different heart rate
-* Add argument `method` in `find_outliers()` calling `_find_outliers_standardize()' or `_find_outliers_percentile()'
 
 
 Fixes
--- a/neurokit2/misc/find_outliers.py
+++ b/neurokit2/misc/find_outliers.py
@@ -4,11 +4,7 @@ import scipy
 from ..stats import standardize
 
 
-def find_outliers(data, 
-                  exclude=0.05, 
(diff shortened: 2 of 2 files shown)


DominiqueMakowski:
Thanks @danibene, sorry I was on holidays, I will review this and merge asap :)

DominiqueMakowski:
@danibene I checked and somewhat simplified the function, as I was not sure about all the various arguments (such as the range multiplier and the difference between the percentile range and threshold). 

I tried to unify the arguments so that everything is controlled using `exclude`, and added examples. Can you check and let me know if there is any particular behaviour that is impossible to achieve using the current implementation? thanks

danibene:
Thanks @DominiqueMakowski ! It looks a lot cleaner now :-) 

One small comment about lines 92-93, since exclude is not an array/list/tuple, there should be no indexing right?
```
            right = np.percentile(data, (1 - (exclude[1] / 2)) * 100)
            left = np.percentile(data, (exclude[0] / 2) * 100)
```

Without the extra arguments (e.g. the range multiplier), we can only have the threshold based directly on the percentile rather than, for example, 1.5*interquartile range + the 75th percentile (as in the "quartiles" method in [MATLAB's isoutlier](https://www.mathworks.com/help/matlab/ref/isoutlier.html#bvlllts-method)).

DominiqueMakowski:
> there should be no indexing right?

good catch!

> rather than, for example, 1.5*interquartile range + the 75th percentile (as in the "quartiles" method in [MATLAB's isoutlier](https://www.mathworks.com/help/matlab/ref/isoutlier.html#bvlllts-method)).

mmh I'd say for now it's better to keep the function fairly generic; this method seems to be somewhat very *matlab-esque* and oddly specific. If users want to mimic this behaviour they can probably do it on their side, as it's not too hard to do. 

If that's okay with you I'll merge :)

danibene:
> this method seems to be somewhat very matlab-esque and oddly specific. If users want to mimic this behaviour they can probably do it on their side, as it's not too hard to do.

I have seen the 1.5*IQR rule outside of MATLAB (e.g. https://online.stat.psu.edu/stat200/lesson/3/3.2), but keeping things simple for now makes sense to me. 

> If that's okay with you I'll merge :)

Yes please do! Thank you for your help with this :-)

DominiqueMakowski:
> I have seen the 1.5*IQR rule outside of MATLAB

Thanks, I didn't know that. Well if in the future we see that there's demand for this method we can always think about adding it in :)

Merging now, thanks again!

Issue #650, opened 2022-06-03
Adapt signal_fixpeaks to deal with larger gaps in data


danibene:
**Describe the solution you'd like**
I noticed that the "neurokit" method of signal_fixpeaks does not work well with larger gaps in the data, sometimes leading to negative indices or getting stuck in a loop: https://github.com/danibene/fixpeaks_with_large_gaps/blob/main/example_signal_with_large_gaps.ipynb

**How could we do it?**
I think there are a few possible ways to go about dealing with larger gaps (not mutually exclusive):
1. Adapting the number of NaNs depending on the interval size rather than always inserting 2 NaNs for large intervals
2. Interpolating the peaks directly rather than the intervals
3. Providing a warning if there are negative indices & breaking the loop after a certain number of iterations

If you're interested in any of the above, I'm happy to try my hand & open a PR!

DominiqueMakowski:
Fixing peaks is a tricky issue, but there's definitely room for improvement!
n°3 should be a starting point, then I don't know about what's best between 1 and 2... but please do open a PR so we can check. 

Would be also nice to add an example / study explaining and benchmarking these methods

danibene:
> Fixing peaks is a tricky issue
Indeed!

I've opened the PR for now with the 3 changes I mentioned above, it's a work in progress but I thought I should get your input before overcomplicating things: https://github.com/neuropsychology/NeuroKit/pull/647 

[The new version is able to fix ECG peaks](https://github.com/danibene/fixpeaks_with_large_gaps/blob/main/example_ECG_signal_distored_nk_mod.ipynb 
) for data [where the old version got stuck in a loop](https://github.com/danibene/fixpeaks_with_large_gaps/blob/main/example_ECG_signal_distored_nk_orig.ipynb).

I still see a couple of issues though:
- there is no extrapolation in case there is a NaN at the final index (so when interpolating the peaks directly, the [index stays constant rather than increasing](https://github.com/danibene/fixpeaks_with_large_gaps/blob/main/example_signal_with_NaN_at_end.ipynb)) 
&rarr; modify the signal_interpolate function to extrapolate? 
&rarr; could remove NaNs at the end?
- sometimes the indices are not negative, but still [out of order](https://github.com/danibene/fixpeaks_with_large_gaps/blob/main/example_signal_with_NaN_at_end.ipynb)
&rarr; provide warning for peak indices that are not increasing? 
[...]

PR #651, opened 2022-06-06
[Feature] Adapt fixpeaks for large gaps

danibene:
# Description

Adapt the neurokit method of fixing peaks for data with larger gaps.

# Proposed Changes

I modified the `signal_fixpeaks()` function as follows (see https://github.com/neuropsychology/NeuroKit/issues/650):
1. Adapting the number of NaNs depending on the interval size rather than always inserting 2 NaNs for large intervals
2. Interpolating the peaks directly rather than the intervals
3. Providing a warning if there are negative indices & breaking the loop after a certain number of iterations

# Checklist
  - However I am still confused about which version current changes go under


DominiqueMakowski:
also calling for the opinion of @JanCBrammer the master of peaks detection :D

JanCBrammer:
@danibene, @DominiqueMakowski, trying to have a look this week (I'll let you know if I don't manage to find time).

JanCBrammer:
@danibene, @DominiqueMakowski, I noticed that we don't have tests that cover the `signal_fixpeaks(method="neurokit")`, we only test the default "kubios" method (as far as I can tell):

https://github.com/neuropsychology/NeuroKit/blob/master/tests/tests_signal_fixpeaks.py
https://app.codecov.io/gh/neuropsychology/NeuroKit/blob/master/neurokit2/signal/signal_fixpeaks.py

@danibene would you mind writing tests for the "neurokit" method, since the PR affects this method? In #650 you already wrote very helpful manual tests in the notebooks. However, it would be even better for the codebase if we'd have these tests run automatically as part of the test suite.

The changes look ok to me otherwise other than a few nitpicky things.

DominiqueMakowski:
@JanCBrammer nitpicking's good! give us what u got ☺️ 

For me I always worry about exposing too much new arguments, to keep the best balance between good-default and simple API and control and flexibility

Also, I'd love to see some of your notebook material from #650 in an [example](https://github.com/neuropsychology/NeuroKit/tree/master/docs/examples) going through peak detection assessment and fixing. Since neurokit doesn't have a GUI to manually fix them, a guide exposing the algorithmic options would be super helpful I think

danibene:
@JanCBrammer I tried adding "neurokit" as an input to one of the test functions, and it won't stop running, so I'm wondering if there is still an infinite loop that isn't caught by the "iterations_max" argument. But since I've never used pytest (or any testing framework) before, it could just be me doing something incorrectly there. In case you'd like to take a look, I created a new branch with the changes: https://github.com/danibene/NeuroKit/tree/feature/adapt_fixpeaks_for_large_gaps_tests

@DominiqueMakowski :

> @JanCBrammer nitpicking's good! give us what u got ☺️

Second this! I'm still learning so nitpicking is extra appreciated :-)

> For me I always worry about exposing too much new arguments, to keep the best balance between good-default and simple API and control and flexibility

I see your point, happy to remove any arguments, but then I wouldn't be sure whether to replace the default in case current users are relying on the current version & it breaks their code?

> Also, I'd love to see some of your notebook material from https://github.com/neuropsychology/NeuroKit/issues/650 in an [example](https://github.com/neuropsychology/NeuroKit/tree/master/docs/examples) going through peak detection assessment and fixing. Since neurokit doesn't have a GUI to manually fix them, a guide exposing the algorithmic options would be super helpful I think

For sure, will do once I debug the infinite loop issue!

JanCBrammer:
@danibene, I'll push some example tests to `danibene:feature/adapt_fixpeaks_for_large_gaps` tomorrow (German time).

JanCBrammer:
@danibene, regarding the tests, have a look at 1901d2c30a01bb720a1078d846752cbab0bd7047. Looking at your notebooks and #650, it looks like the "neurokit" method of `signal_fixpeaks()` should meet at least three criteria:

The corrected peak-indices should...
1. ...be strictly increasing.
2. ...not contain duplicates.
3. ...not be negative.

I tried to express these criteria as tests by pasting the code from your notebooks into test functions.

In the root of the NeuroKit repository, you can now run `pytest -v tests/tests_signal_fixpeaks.py::test_neurokit_method_returns_strictly_increasing_indices`.
You can replace the function name after the `::` with the name of another function in the `test_signal_fixpeaks` module.
If you want to run all tests in `test_signal_fixpeaks`, leave out the `::test_...` part.
Let me know if there are problems with this, it might be that you have to install the `pytest-coverage` plugin.
Also let me know if you have questions regarding the implementation of the tests (e.g., fixtures, parameteriztion etc.).

According to the test output, criteria 1. and 2. aren't met. While working on the code you now have quick and convenient feedback: you can make a change, run the tests, and see if you're going in a good direction. Some people call this [test driven development](https://en.wikipedia.org/wiki/Test-driven_development). I.e., you're writing a failing test and then try to make it pass.

I think it makes most sense if I review the implementation once the tests are passing, I assume that there will be changes along the way, so its not super-useful if I give detailed feedback now.

Regarding the exposure of the additional parameters, I also think that there shouldn't be too many. I think it'd be good to always determine the number of NaNs based on the interval-size, i.e., getting rid of the `n_nan_large_interval` parameter.
Also, I think it's best to remove the `interpolate_on_peaks` parameter and hard-code the option that works best.
Finally, regarding the `iterations_max` parameter: couldn't it be replaced by a more adaptive criterion? For example, stop iterating once any of the interpolated intervals gets smaller than some threshold?

danibene:
Thank you @JanCBrammer that was all very helpful!

I changed the test functions as follows:
- `n_nan_large_interval` is `None` (to replicate how the function should behave once that parameter is removed) 
- I also deleted a peak at the end of the test signal to demonstrate the problem I mentioned in https://github.com/neuropsychology/NeuroKit/issues/650 that the indices at the end stay constant, causing the tests to fail.

I see two possible ways to address this:
1. Modify the [signal_interpolate](https://neurokit2.readthedocs.io/en/latest/_modules/neurokit2/signal/signal_interpolate.html) function to extrapolate and use that instead of Pandas for interpolation.
2. If there are NaNs at the beginning or end, remove them. This would mean that there will be missed peaks at the beginning/end, which as far as I know would not be problematic for computing interbeat intervals, but might be for some other use case?

Concerning the `iterations_max` parameter, I agree that some kind of adaptive criterion would make more sense
> stop iterating once any of the interpolated intervals gets smaller than some threshold

The natural answer for me here would be before the minimum interval reaches `interval_min`, but this could already happen in the first loop. Maybe we should repeat the `_remove_small` step in that case, or at least raise a warning?

JanCBrammer:
@danibene

# Extrapolation

> I see two possible ways to address this:
> 
> 1. Modify the [signal_interpolate](https://neurokit2.readthedocs.io/en/latest/_modules/neurokit2/signal/signal_interpolate.html) function to extrapolate and use that instead of Pandas for interpolation.

I was wondering anyway why the Pandas interpolation is used here instead of our own `signal_interpolate()`. As far as I now `signal_interpolate()` does extrapolate at the signal's edges, doing what we want? 
https://github.com/neuropsychology/NeuroKit/blob/277f614ee3984647d71ba31aadffea97766aeb0f/neurokit2/signal/signal_interpolate.py#L18-L19

# Stopping criterion
> The natural answer for me here would be before the minimum interval reaches `interval_min`, but this could already happen in the first loop.

Good point. How about using some kind of aggregate statistic ? We could compute that statistic on the artifact-free data (`s_af`), then compute it on each iteration on the artifact data (`s_a`), and stop iterating once `s_a` deviates too much from `s_af`? Could be costly to compute `s_a` on each iteration, but might be worth it?

danibene:
@JanCBrammer 
> I was wondering anyway why the Pandas interpolation is used here instead of our own signal_interpolate(). As far as I now signal_interpolate() does extrapolate at the signal's edges, doing what we want?

[signal_interpolate()](https://neurokit2.readthedocs.io/en/latest/_modules/neurokit2/signal/signal_interpolate.html) passed `fill_value=([y_values[0]], [y_values[-1]])` to [scipy.interpolate.interp1d()](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.interp1d.html#scipy.interpolate.interp1d), so there were repeated values outside the data range. After I [added `fill_value` as an argument in signal_interpolate() and set it to `"extrapolate"` in signal_fixpeaks()](https://github.com/neuropsychology/NeuroKit/pull/651/commits/bc70af812ea07f49bfc569a174e15ebf35ab4d51), all the current tests passed. 

What do you think of this solution? I'm unsure about whether we should keep the default interpolation method (`"quadratic"` is the default of [signal_interpolate()](https://neurokit2.readthedocs.io/en/latest/_modules/neurokit2/signal/signal_interpolate.html) vs. `"linear"` is the default in [Pandas](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.interpolate.html)).

> Good point. How about using some kind of aggregate statistic ? We could compute that statistic on the artifact-free data (s_af), then compute it on each iteration on the artifact data (s_a), and stop iterating once s_a deviates too much from s_af? Could be costly to compute s_a on each iteration, but might be worth it?

Very interesting idea! Did you have any specific statistic in mind (e.g. RMSSD)? Also, do you have any thoughts on how to determine the threshold for the acceptable deviation (empirically with distorted data, theoretically)?

JanCBrammer:
@danibene, l had a closer look at how the NeuroKit method handles large gaps in peak-locations and I suspect that the problem can be fixed by getting rid of the while loop
https://github.com/neuropsychology/NeuroKit/blob/277f614ee3984647d71ba31aadffea97766aeb0f/neurokit2/signal/signal_fixpeaks.py#L589

and the interpolation
https://github.com/neuropsychology/NeuroKit/blob/277f614ee3984647d71ba31aadffea97766aeb0f/neurokit2/signal/signal_fixpeaks.py#L618

When you opened the PR, you already described a good, simple, alternative approach:

> 1. Adapting the number of NaNs depending on the interval size rather than always inserting 2 NaNs for large intervals
> 2. Interpolating the peaks directly rather than the intervals

The idea would be to simply fill the gaps in peak locations with equally spaced peak locations (in one go, that is, not iteratively in a loop). The spacing between the "filled-in" peak locations would be equal to the mean inter-beat-interval in the clean data. 
In other words, instead of adding the NaNs and then using an interpolation function to fill the NaNs with peak locations, why not add the peak locations directly?

Then there would be no need for iterating (i.e., while loop) and the stopping criterion.

Let me know if I should describe what I mean in more detail, not sure how much sense it makes atm?

danibene:
@JanCBrammer 

> When you opened the PR, you already described a good, simple, alternative approach:
> 
> > 1. Adapting the number of NaNs depending on the interval size rather than always inserting 2 NaNs for large intervals
> > 2. Interpolating the peaks directly rather than the intervals
> 
> The idea would be to simply fill the gaps in peak locations with equally spaced peak locations (in one go, that is, not iteratively in a loop). The spacing between the "filled-in" peak locations would be equal to the mean inter-beat-interval in the clean data. In other words, instead of adding the NaNs and then using an interpolation function to fill the NaNs with peak locations, why not add the peak locations directly?
> 
> Then there would be no need for iterating (i.e., while loop) and the stopping criterion.
> 
> Let me know if I should describe what I mean in more detail, not sure how much sense it makes atm?

Could you confirm whether what you mean is equivalent to the current version in my branch when `interpolate_on_peaks=True`?
https://github.com/danibene/NeuroKit/blob/bc70af812ea07f49bfc569a174e15ebf35ab4d51/neurokit2/signal/signal_fixpeaks.py#L757-L763
(couldn't figure out how to embed the lines in this comment, you will have to click 😢)

If you didn't mean what is done in the current version, could you clarify what you would do differently? I'm not sure how we would fill in the peak locations without adding NaNs and then interpolating.

Thank you :-)

JanCBrammer:
> Could you confirm whether what you mean is equivalent to the current version in my branch when `interpolate_on_peaks=True`?
> https://github.com/danibene/NeuroKit/blob/bc70af812ea07f49bfc569a174e15ebf35ab4d51/neurokit2/signal/signal_fixpeaks.py#L757-L763

That's what I mean 👍 If you determine the number of NaNs like this it should be possible to get rid of the while loop https://github.com/neuropsychology/NeuroKit/blob/277f614ee3984647d71ba31aadffea97766aeb0f/neurokit2/signal/signal_fixpeaks.py#L589

Let's also go with your idea to make `interpolate_on_peaks` the default, i.e., removing the parameter. Then we can also get rid of the https://github.com/neuropsychology/NeuroKit/blob/277f614ee3984647d71ba31aadffea97766aeb0f/neurokit2/signal/signal_fixpeaks.py#L626 helper function.


> I'm not sure how we would fill in the peak locations without adding NaNs and then interpolating.

Let's interpolate using the NaNs for now.

DominiqueMakowski:
> I was wondering anyway why the Pandas interpolation is used here instead of our own signal_interpolate()

Probably it was coded before the implementation of signal_interpolate 🤷? But feel free to change to whatever makes more sense!

danibene:
@JanCBrammer 

> Speaking of `signal_interpolate()`, @danibene, since the changes to the interpolation are somewhat extensive, could you make a separate PR for them? This would keep the current PR more concise and focused on the changes to `signal_fixpeaks(). 

Done: https://github.com/neuropsychology/NeuroKit/pull/666

> You could keep the Pandas interpolation for this PR.

If I do that, I'd still have to decide on what to do about the fact that the Pandas interpolation does not extrapolate, which causes the tests to fail if there are NaNs at the edge. There would be this option still though:

>    If there are NaNs at the beginning or end, remove them. This would mean that there will be missed peaks at the beginning/end, which as far as I know would not be problematic for computing interbeat intervals, but might be for some other use case?

neurokit2/signal/signal_fixpeaks.py
```diff
@@ -563,67 +624,92 @@ def _plot_artifacts_lipponen2019(artifacts, info):
...
-            interval = standardize(interval, robust=robust)
-            peaks, continue_loop = _interpolate_missing(
-                peaks, interval, relative_interval_max, sampling_rate
-            )
-
+    if interval_max is not None:
+        interval = signal_period(
+            peaks, sampling_rate=sampling_rate, desired_length=None
+        )
+        peaks, continue_loop = _interpolate_missing(
```
JanCBrammer, line 661: With the while loop gone, `_interpolate_missing()` doesn't have to return `continue_loop` anymore.

neurokit2/signal/signal_interpolate.py
```diff
@@ -3,7 +3,9 @@
 import scipy.interpolate
 
 
-def signal_interpolate(x_values, y_values, x_new=None, method="quadratic"):
```
JanCBrammer, line 6: Since there's a separate PR for the changes to `signal_interpolate()` now, could you revert all changes to this file (including formatting) such that it doesn't show up in this PR?

neurokit2/signal/signal_fixpeaks.py
danibene: Thanks for catching that!

neurokit2/signal/signal_fixpeaks.py
```diff
@@ -563,67 +594,89 @@ def _plot_artifacts_lipponen2019(artifacts, info):
...
-    peaks = np.insert(
-        peaks, outliers_loc, peaks_corrected[outliers_loc + np.arange(len(outliers_loc))]
-    )
-    return peaks, True
+    # go through the outliers starting with the highest indices
+    # so that the indices of the other outliers are not moved when
+    # unknown intervas are inserted 
+    for loc in np.flip(outliers_loc):
+        # compute number of NaNs to insert based on the mean interval
+        n_nan = int(interval[loc] / np.nanmean(interval_without_outliers))
```
JanCBrammer, line 671: One final nitpick 🙈 `np.nanmean(interval_without_outliers)` won't change throughout the iterations of the for loop. I.e., it's a loop-invariant that's unnecessarily re-computed on every iteration. You could assign it to a variable before you enter the loop, and then use the variable inside the loop (i.e., "hoist the loop-invariant").

JanCBrammer (approved):

neurokit2/signal/signal_fixpeaks.py
danibene: Yes good point, and thanks for teaching me the vocab :D

DominiqueMakowski:
I fixed the docs (the remaining one for readthedocs can be disregarded). I can merge whenever you want!

(and thanks again @danibene @JanCBrammer for this thorough and super-clean work&review)

JanCBrammer:
Thanks @DominiqueMakowski, good to merge from my side!

danibene:
Thank you both, good to merge from my side too :)

PR #666, opened 2022-06-22
[Feature] Modify signal_interpolate for input with NaNs and extrapolation

danibene:
# Description

Modify the `signal_interpolate()` function to accept a single array containing NaNs as input and to extrapolate. 

(See https://github.com/neuropsychology/NeuroKit/pull/651 for original motivation for doing so)

# Proposed Changes

I modified the `signal_interpolate()` function as follows:
1. Accepting a single array containing NaNs as input (either `x_values` or `y_values` can be `None`), for interpolation of missing values like in [Pandas](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.interpolate.html)
2. Adding `fill_value` as an argument to be passed to the [Scipy interpolate function](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.interp1d.html), such that it can be set to "extrapolate"

# Checklist


DominiqueMakowski:
What's the status here is it good for review/merge or still wip?

danibene:
@DominiqueMakowski @JanCBrammer I think it's good for review, but let me know if I'm missing anything!

JanCBrammer:
Will try to have a look this week. Please @ me here in case you don't hear back.

neurokit2/signal/signal_interpolate.py
```diff
@@ -29,27 +29,28 @@ def signal_interpolate(x_values, y_values, x_new=None, method="quadratic"):
         order of the spline interpolator to use.
         See `here <https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.
         PchipInterpolator.html>`_ for details on the ``"monotone_cubic"`` method.
-
+    fill_value : float or tuple or str
+        If a ndarray (or float), this value will be used to fill in for
+        requested points outside of the data range.
+        If a two-element tuple, then the first element is used as a fill value
+        for x_new < x[0] and the second element is used for x_new > x[-1].
+        If “extrapolate”, then points outside the data range will be extrapolated.
```
DominiqueMakowski, line 37:
```suggestion
        If "extrapolate", then points outside the data range will be extrapolated.
```

DominiqueMakowski:
@JanCBrammer

JanCBrammer:
@danibene, @DominiqueMakowski, I recall that settling on an API for `signal_interpolate()` wasn't straightforward: #247. This is why I'm hesitant to change it.

# Pandas-style NaN interpolation
The NaN interpolation is a rather big modification of the API: going from mandatory `x_values` and `y_values` to being able to omitting one of them. This modification adds complexity and overhead and I'm not sure it's worth it. What do you think @DominiqueMakowski.

# custom extrapolation values
I'm not opposed to adding the `fill_value` parameter. Ideally it should be applied to the `monotone_cubic` method as well.

DominiqueMakowski:
@Tam-Pham tell us what you think so we can merge or not and release 2.1

DominiqueMakowski:
@Tam-Pham *bump*

DominiqueMakowski:
I'm not sure what to do here, I don't have a clear idea of the risks / benefits balance, and @JanCBrammer comment makes me cautious. Is it indeed a change that will benefit many users and carries no risk?

danibene:
@DominiqueMakowski 

> I'm not sure what to do here, I don't have a clear idea of the risks / benefits balance, and @JanCBrammer comment makes me cautious. Is it indeed a change that will benefit many users and carries no risk?

It's hard for me to say - maybe an alternative could be that I remove the change that @JanCBrammer was hesitant about, i.e.: 

> Accepting a single array containing NaNs as input (either x_values or y_values can be None), for interpolation of missing values like in [Pandas](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.interpolate.html)

and I could open a separate PR for a separate function that calls `signal_interpolate()` and takes only one input array? E.g. `signal_interpolate_nan()`? Then this separate function could eventually be used in `signal_fixpeaks()` instead of Pandas.

DominiqueMakowski:
So if I understand, we want to add the possibility of interpolating arrays with nan pandas-style to bypass the usage of pandas and allow more control over it?

To do so, we have modified signal_interpolate, that previously required x_values *and* y_values, now if only one is passed, and has Nans, it will interpolate them am I right?

danibene:
@DominiqueMakowski 

> So if I understand, we want to add the possibility of interpolating arrays with nan pandas-style to bypass the usage of pandas and allow more control over it? 

> To do so, we have modified signal_interpolate, that previously required x_values and y_values, now if only one is passed, and has Nans, it will interpolate them am I right?

Yes that's all correct! 

Do you think we should keep this addition in this PR or just the changes that allow extrapolation of points outside of the data range?

DominiqueMakowski:
i think it's fine, perhaps to make it clearer we can have an internal function called `_signal_interpolate_nan()` and if y_values is None and x_values contains NaN, we directly call this function (so that it's internally "separated" from the rest of the regular interpolation behavior)

I guess as long as it doesn't break the current behavior it's good

danibene:
@DominiqueMakowski 

> i think it's fine, perhaps to make it clearer we can have an internal function called _signal_interpolate_nan() and if y_values is None and x_values contains NaN, we directly call this function (so that it's internally "separated" from the rest of the regular interpolation behavior)

Done, let me know if you'd like me to change anything!

Also, I changed an unrelated line in https://github.com/neuropsychology/NeuroKit/pull/666/commits/cd020535b816d8e8af246e237fdcf5738dfa8f8d
https://github.com/neuropsychology/NeuroKit/blob/277f614ee3984647d71ba31aadffea97766aeb0f/neurokit2/signal/signal_interpolate.py#L80-L81
It returned the original y values if the length of the new x values matched the length of the original x values, but the length can still match without the values matching, so instead I changed it to check that all the values match. 

[Here](https://gist.github.com/danibene/ca1ee78016b96d6c030218dde09d4088) is an example of the difference in behavior in case the x values don't match but their length does

DominiqueMakowski:
Thanks a ton Dani for all that work!!

PR #677, opened 2022-07-14
[Feature] Gaussian-Copula Mutual Information

DominiqueMakowski:
- [ ] Try re-implementing GCMI for signal processing.

There is an open-source Python implementation in https://github.com/robince/gcmi, but it gives unexpected results, below is a comparison of GCMI ("MI6") vs. other methods. It doesn't follow the same pattern at all...

```python
import numpy as np
import pandas as pd

x = np.random.normal(size=400)
y = x**2

data = pd.DataFrame()
for level in np.linspace(0.01, 3, 200):
    noise = np.random.normal(scale=level, size=400)
    rez = pd.DataFrame({"Noise": [level]})
    rez["MI1"] = nk.mutual_information(x, y + noise, method="varoquaux", sigma=1)
    rez["MI2"] = nk.mutual_information(x, y + noise, method="varoquaux", sigma=0)
    rez["MI3"] = nk.mutual_information(x, y + noise, method="nolitsa")
    rez["MI4"] = nk.mutual_information(x, y + noise, method="knn")
    rez["MI5"] = nk.mutual_information(x, y + noise, method="max")
    rez["MI6"] = nk.mutual_information(x, y + noise, method="gc")
    data = pd.concat([data, rez], axis=0)

# Rescale on the same range for visualization purposes
data["MI1"] = nk.rescale(data["MI1"])
data["MI2"] = nk.rescale(data["MI2"])
data["MI3"] = nk.rescale(data["MI3"])
data["MI4"] = nk.rescale(data["MI4"])
data["MI5"] = nk.rescale(data["MI5"])
data["MI6"] = nk.rescale(data["MI6"])

data.plot(x="Noise", y=["MI1", "MI2", "MI3", "MI4", "MI5", "MI6"], kind="line")
```

image

### DIFF
--- a/neurokit2/complexity/information_mutual.py
+++ b/neurokit2/complexity/information_mutual.py
@@ -66,6 +66,7 @@ def mutual_information(x, y, method="varoquaux", bins="default", **kwargs):
       nk.mutual_information(x, y, method="nolitsa")
       nk.mutual_information(x, y, method="knn")
       nk.mutual_information(x, y, method="max")
+      nk.mutual_information(x, y, method="gc")
 
     **Example 2**: Method comparison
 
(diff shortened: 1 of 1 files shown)


robince:
I think your signal is `y=x**2 + noise`. GCMI won't detect this sort of symmetric effect, it can detect only monotonic effects (it is like a rank correlation). You could try adding in abs(x) as a second dimension of x. 

Also I apprecaite your interest but GCMI is licensed under the GPL so code can't be reused and released under MIT.

PR #678, opened 2022-07-15
[Fix] ppg_plot() was not returning matplotlib.Figure

luiseduve:
Hi...

# Description

I needed to store the results of `ppg_plot()` in a file on disk (_easier to inspect than doing it manually in the notebook_). However, the function did not return the object `matplotlib.figure` so I could use `fig.savefig([path])`, [even though the documentation says that it returns a figure](https://neuropsychology.github.io/NeuroKit/functions/ppg.html#ppg-plot).

If the PR is accepted, I can check similar behavior for the other variables.

# Proposed Changes

- Return the object `matplotlib.figure.Figure` for `ppg_plot()` instead of just plotting the image on the notebook.
- [Docs] _Minor edit_: Added description about feature LFHF returned by `hrv_frequency()`, it was missing.

# Checklist


DominiqueMakowski:
Thanks for that PR, unfortunately, we recently changed that behaviour on behaviour to *not* return the figure, as it was causing some issues, among which double figures showing up in the console and inconsistent return type which is not a good programming practice (that said, it's possible that documentation has to be adjusted in some places!)

In principle thanks to the arcanes of matplotlib you can still recover the figure and save it to disk as follows:

```python
import neurokit2 as nk

# Generate 15 seconds of PPG signal (recorded at 250 samples / second)
ppg = nk.ppg_simulate(duration=15, sampling_rate=250, heart_rate=70, random_state=333)

# Process it
signals, info = nk.ppg_process(ppg, sampling_rate=250)

# Visualize the processing
nk.ppg_plot(signals, sampling_rate=250)

# Save it
plt.tight_layout()
plt.savefig("ppg.png", dpi=300)
```

Or even 

```python
fig = plt.gcf()
fig.set_size_inches(10, 6, forward=True)
fig.savefig("ppg2.png", dpi=300)
```

Let us know if it works, and if it does, you can edit your PR to just leave the documentation fix :)

luiseduve:
Thanks for your feedback 😁. Your example code to save the plots worked very well!
I have updated the docs related with plot functions, deleting the _Return_ and adding an example to save figures to PNG files.

luiseduve:
Apparently there is an error building the new docs induced by my change in `ecg_plot()`. This is the first time doing a PR and working with sphinx, so I am not completely sure what went wrong 😅. How can I reproduce locally the _Documentation check_ before doing a new commit?

DominiqueMakowski:
I fixed the issues, merging now :) thanks!

PR #679, opened 2022-07-20
[Fix] Calculation of number of NaNs in signal_fixpeaks()

danibene:
# Description

This PR aims to improve the previous PR: https://github.com/neuropsychology/NeuroKit/pull/651
See also the related issue: https://github.com/neuropsychology/NeuroKit/issues/650

# Proposed Changes

I changed the `neurokit` method of the `signal_fixpeaks()` function so that the number of NaNs inserted is computed as follows:
1. Rounding rather than taking the floor when getting an integer value for the number of NaNs (calculated by dividing large gaps by the mean interval)
2. Calculating the mean interval based on the original interval values and not the standardized interval values in the case that `relative_interval_max` is given.

# Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/neuropsychology/NeuroKit/blob/master/.github/CONTRIBUTING.rst#structure-and-code) file.
- [x] My PR is targeted at the **dev branch** (and not towards the master branch).
- [x] I ran the [CODE CHECKS](https://github.com/neuropsychology/NeuroKit/blob/master/.github/CONTRIBUTING.rst#run-code-checks) on the files I added or modified and fixed the errors.
  I got the following error, but I'm just going to ignore it if that's okay, since I don't think I changed anything that should cause this? 
  `astroid.exceptions.TooManyLevelsError: Relative import with too many levels (2) for module 'signal_fixpeaks'` Could also be the fact that I have a file called `signal_fixpeaks.py` in a different folder.
- [ ] I have added the newly added features to **News.rst** (if applicable)

Issue #671, opened 2022-07-07
Interpolation in hrv_frequency()


danibene:
**Question and context**

Hello! 

I would like to try interpolating R-R intervals at different sampling frequencies to see how that affects the frequency-domain HRV features, but I am confused about how the R-R intervals are interpolated in `hrv_frequency()`:
https://github.com/neuropsychology/NeuroKit/blob/277f614ee3984647d71ba31aadffea97766aeb0f/neurokit2/hrv/hrv_frequency.py#L162-L165

Looking at `_hrv_get_rri()`, I am unsure whether `sampling_rate` is the appropriate variable to change here: 

https://github.com/neuropsychology/NeuroKit/blob/277f614ee3984647d71ba31aadffea97766aeb0f/neurokit2/hrv/hrv_utils.py#L8-L29

My questions:
1. Where is the sampling frequency of the R-R interval interpolation set? It seems to me like `desired_length` will be the same regardless of the value of `sampling_rate`.
2. Why is the minimum sampling frequency set to 10 Hz? I was wondering because I've seen lower values in the literature for interpolation of R-R intervals (e.g. [4 Hz](https://ieeexplore.ieee.org/abstract/document/6090371?casa_token=o3JfJXjUsCgAAAAA:pNq6QiTQkMuQgV_c68ufrYtR8e0SDlCOrmhaduc1eB9k_W1byv6sz1QspeYf5ofcxMMTnD2rjw)) - if I understand correctly that `sampling_rate` is referring to the frequency at which the R-R intervals should be interpolated.

Thank you :)

DominiqueMakowski:
Hey Dani, 

1. I am not entirely sure (tagging @Tam-Pham). The thing is that we need to know the sampling rate regardless of the interpolation purposes to make sense of the values. For instance, RRis that [2000, 1500, 1750] and [100, 75, 87.5] might be the same (the first one is differences at 1000Hz, the second one at 50Hz), so we need to put them on the same scale (the y-axis). Then, the rate of interpolation is something different. I'm not sure it helps but that's what it makes me think
2. I'm not sure either. The minimum sampling rate *might* have to do with the power spectrum that is computed, since too coarse signals might lead to errors there. I saw 4 Hz for sampled heart rates (some "smart"watch & wearables only provide that), but that's definitely too low and too coarse for PSD estimation, better to resample to higher freqs.

Let us know your thoughts

danibene:
[...]

PR #680, opened 2022-07-20
[Feature] Option to change interpolation rate of interbeat intervals for HRV frequency

danibene:
# Description

This PR aims to enable custom interpolation rates of interbeat intervals. See https://github.com/neuropsychology/NeuroKit/issues/671

# Proposed Changes

I changed the `_hrv_get_rri()` function so that:
1. There is a new argument `interpolation_rate` instead of always setting the `sampling_rate` of the raw heartbeat data as the interpolation rate of the interbeat intervals.
2. The default is 1000 Hz (up for discussion?).
3. There is no more minimum requirement of 10 Hz (up for discussion?).

# Checklist
  Got the same error as in https://github.com/neuropsychology/NeuroKit/pull/679 that I will continue to ignore if that's okay: `astroid.exceptions.TooManyLevelsError: Relative import with too many levels (2) for module 'hrv_utils'`
  I am still confused about where these should be listed


danibene:
Regarding the failing checks (sorry, didn't notice until I checked my email):

For [README_examples](https://github.com/neuropsychology/NeuroKit/pull/680/checks#step:5:21):
I have changed the line causing the error, `interpolated[int(x_values[-1]) :] = interpolated[int(x_values[-1])]`, in https://github.com/neuropsychology/NeuroKit/pull/666 - maybe it makes sense to decide on whether those changes will be merged before addressing the current PR?

For [tests_hrv](https://github.com/neuropsychology/NeuroKit/runs/7435027575?check_suite_focus=true), I'm having trouble interpreting the output: `tests/tests_hrv.py::test_hrv_frequency ERROR: InvocationError for command /home/runner/work/NeuroKit/NeuroKit/.tox/py/bin/pytest tests neurokit2/bio neurokit2/complexity neurokit2/data neurokit2/ecg neurokit2/eda neurokit2/eeg neurokit2/emg neurokit2/eog neurokit2/epochs neurokit2/events neurokit2/hrv neurokit2/microstates neurokit2/misc neurokit2/ppg neurokit2/rsp neurokit2/signal neurokit2/stats -v (exited with code -9 (SIGKILL)) (exited with code -9)` 
Would it be possible for someone to help me with this?

Thanks :)

DominiqueMakowski:
It seems like this is breaking several tests, we need to evaluate that and update the tests if need be

DominiqueMakowski:
I'm half-convinced ab out the 4 Hz by default, it just seems really low to get frequency decomposition :/

danibene:
@DominiqueMakowski 
> I'm half-convinced ab out the 4 Hz by default, it just seems really low to get frequency decomposition :/

How about a default of 7 or 10 Hz? 

From https://pubmed.ncbi.nlm.nih.gov/15825865/

"Most of the papers in the field of
HRV report on the use of resampling rates between 2 and 4 Hz (although 1 and 10 Hz have also been used) [2], [3]. Choosing a resampling rate of 7 Hz for the RR tachogram has the advantage that HR <= 210 BPM can be resolved (see Section III-A2 on the average Nyquist criterion)."

DominiqueMakowski:
That sounds better, though it's still unclear to me why not resample at higher rates, I mean the computational overhead is quite and higher rates could accommodate users that work with animals (that can have super high bpms)

I mean aside from "it's not what kubios does", it's not clear to me the disadvantages of resampling at 100+ Hz... do you know of any? (I'm sorry for blocking here but I wanna make sure we make the best decisions for our defaults ^^)

danibene:
@DominiqueMakowski 

> (I'm sorry for blocking here but I wanna make sure we make the best decisions for our defaults ^^)

No worries, I appreciate the due diligence :) 

> I mean aside from "it's not what kubios does", it's not clear to me the disadvantages of resampling at 100+ Hz... do you know of any? 

Not really, I mainly thought for reproducibility's sake it's better to keep it within the range that is used in the literature, as well in other software beyond Kubios e.g.
- 4 Hz: https://github.com/PGomes92/pyhrv/blob/master/pyhrv/frequency_domain.py#L161
- 4 Hz: https://github.com/Aura-healthcare/hrv-analysis/blob/master/hrvanalysis/extract_features.py#L218-L220

But I also understand if you'd rather not change it from the current default, and with `interpolation_rate` as a separate argument, the user will be able to specify anyway.

DominiqueMakowski:
<sub>I am pretty sure these two packages you mentioned have 4 Hz just because of kubios though 😁</sub>

Okay, in that case let's put it at 100, which should be enough to cover animal-fast heart rates, and at the same time leads to 10 times shorter signals than 1000 Hz for the analyses (so we gain in efficiency). We can add in the docstring that in order to replicate the results from other software such as Kubios, one can set the argument to 4.

And in a comment right above where we resample, we can summarize our thoughts: We thought to use 4 by default to match Kubios, but it seems too low (especially for some applications with fast heart rates), hence we have put it to 100 by default, but if anybody has any better idea, please do let us know. And we can add a link to this PR :)

DominiqueMakowski:
It seems like this broke `hrv_rsa()`

danibene:
@DominiqueMakowski sorry for not fixing that bug earlier, I think it's good now!

DominiqueMakowski:
looks good, thanks a lot Dani

PR #683, opened 2022-07-22
[Feature] MFDFA: allow for q=0

DominiqueMakowski:
https://github.com/LRydin/MFDFA/issues/33

- [x] Implement *maxFDA*

### DIFF
--- a/neurokit2/complexity/fractal_dfa.py
+++ b/neurokit2/complexity/fractal_dfa.py
@@ -208,7 +208,7 @@ def fractal_dfa(
     q = _sanitize_q(q, multifractal=multifractal)
 
     # Store parameters
-    info = {"scale": scale, "q": q[:, 0]}
+    info = {"scale": scale, "q": q}
 
     # Preprocessing
     if integrate is True:
@@ -229,7 +229,7 @@ def fractal_dfa(
         trends = _fractal_dfa_trends(segments, window, order=order)
 
         # Get local fluctuation
-        fluctuations[i] = _fractal_dfa_fluctuation(segments, trends, multifractal, q)
+        fluctuations[i] = _fractal_dfa_fluctuation(segments, trends, q)
 
     # I would not advise the next part part. I understand the need to remove zeros, but I
     # would instead suggest masking it with numpy.ma masked arrays. Note that
@@ -254,7 +254,7 @@ def fractal_dfa(
 
     # Extract features
     if multifractal is True:
-        info.update(_singularity_spectrum(q=q[:, 0], slopes=slopes))
+        info.update(_singularity_spectrum(q=q, slopes=slopes))
         out = pd.DataFrame(
             {
                 k: v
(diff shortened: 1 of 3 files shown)


DominiqueMakowski:
```python
import matplotlib
import neurokit2 as nk

signal = nk.signal_simulate(duration=10, frequency=[5, 7, 10, 14], noise=0.05)
mfdfa, info = nk.fractal_mfdfa(signal, q=np.linspace(-3, 3, 7), show=True, scale = nk.expspace(50, 500, 20))

fig = matplotlib.pyplot.gcf()
fig.set_size_inches(18.5, 10.5)
```

image


@LRydin I tried to implement it but I suspect that something is off, from the plot as the fluctuation values for when q=0 are weird

neurokit2/complexity/fractal_dfa.py
```diff
@@ -291,6 +291,132 @@ def fractal_dfa(
...
+    q_non0 = q[~is0].reshape(-1, 1)
+
+    # Get the fluctuation function, which is a function of the windows and of q
+    # When q = 2 (i.e., multifractal = False)
+    # The formula is equivalent to np.sqrt(np.mean(var))
+    # And corresponds to the Root Mean Square (RMS)
+    fluctuation = np.float_power(np.mean(np.float_power(var, q_non0 / 2), axis=1), 1 / q_non0.T)
+
+    if np.sum(is0) > 0:
+        fluc0 = np.exp(np.mean(np.log(var)))
```
DominiqueMakowski, line 414: what I did is that I computed the values for q=0 here and then inserted them in the fluctuation list

LRydin: Could you maybe give this a try (sorry for not trying myself, I have to finish something today, I was just puzzled with the disparity in the plot. I think it has to do with a `0.5` that is missing):

```python
fluc0 = np.exp(0.5*np.mean(np.log(var)))
```

DominiqueMakowski: image

ah yes that looks better!

LRydin:
Perfect, the result is correct, you can see that the maximum of the _singularity dimension (D)_ is `1`, and that comes to be when we have `q=0`, so that's perfect :). That was easier than I expected in the end 👍.

DominiqueMakowski:
I added an "experimental" feature that I wanted your opinion on: 

The fluctuations often display a non-linear pattern, and the longer the scales the more the coefficient of the linear slope is biased:

```python
import matplotlib
import neurokit2 as nk

signal = nk.signal_simulate(duration=10, frequency=[5, 7, 10, 14], noise=0.05)
_, _ = nk.fractal_dfa(signal, show=True)
```

image


Intuitively the linear fit seems off.

I added the option to detect the knee of the fluctuation pattern (using the *kneedle* algorithm), and then use that knee as a max. scale before computing the slopes. This results in much closer fits:

```python
_, _ = nk.fractal_dfa(signal, maxdfa=True, show=True)
```

image


For multifractal, I use the max. knee (of the patterns corresponding to different qs) as the limit:



image


image


Does that make sense to you?

DominiqueMakowski:
@LRydin I'll merge this, but still do let me know if you have any thoughts on that when you have time ;) and thanks again for your help!

LRydin:
OMG, I never replied to this! I'm sorry :(! Yes, this makes absolute sense. That plateau that you see on your second image is fairly normal for computer-generated data. To be best of my understanding, i.e., I've never seen a paper discuss this, the plateau appears as a limit to the standard deviation of your entire process, so, putting it simply, (MF)DFA cannot detect changes in the standard deviation because the process `signal` is "bounded" between to values (quotation marks there), so it makes much more sense to extract the flucutation function at a lag smaller than the kink/keen :)

It's great! Sorry I took forever to reply @DominiqueMakowski  :|

DominiqueMakowski:
No worries, thanks a lot @LRydin ;)

PR #693, opened 2022-08-11
[Fix] Fixed failing complexity test

rostro36:
# Description

Because _misc/expspace.py_ was changed in [this commit](https://github.com/neuropsychology/NeuroKit/commit/d68f9ef42357823e23c07242a1db2f616d4fd212), the adjacent test now fails. As also proposed in  #683 , I now also want to just adjust the numbers to what I get in a notebook. Also I linted the whole file.

If there are other (py)tests failing, I also want to clean that up over the weekend.
# Proposed Changes

The interesting changes are on line 40 and 42 with new exact numbers. The rest is just linting; I also changed from a print message to an exception in line 634.

# Checklist

### DIFF
--- a/tests/tests_complexity.py
+++ b/tests/tests_complexity.py
@@ -1,18 +1,23 @@
+import math
+
+from collections.abc import Iterable
+
 import antropy
 import nolds
 import numpy as np
 import pandas as pd
+
 from pyentrp import entropy as pyentrp
+from sklearn.neighbors import KDTree
 
 # import EntropyHub
 import neurokit2 as nk
 
-"""
-For the testing of complexity, we test our implementations against existing and established ones.
-However, some of these other implementations are not really packaged in a way
-SO THAT we can easily import them. Thus, we directly copied their content in this file
-(below the tests).
-"""
+
+# For the testing of complexity, we test our implementations against existing and established ones.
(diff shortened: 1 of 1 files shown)


DominiqueMakowski:
sorry for the delay I just came back from vacations and it's pretty busy!
Thanks a lot for this PR. Overall, I think the test suite in neurokit could be improved, so don''t hesitate if you have any more ideas ☺️

tests/tests_complexity.py
```diff
@@ -1,18 +1,23 @@
+import math
```
DominiqueMakowski, line 1: Do we need that? i don't see any call to `math`

rostro36: Thank you for going over it in detail.

There are two calls to `math` in line 596 and line 598, which did not get changed and do not show in the commit message.

`isort` moved the input to the top, before it was in front of the test with the math calls.

rostro36:
No worries, I am grateful for all the work that you do for this project and did not expect to get answers right away.

I don't have that much experience with testing. There are a couple of other things I want to do first (like adding more RVT measures), but I can look into the test suite afterwards if I have time.

Generally, I think it already is good if there is an automated test suite.

Issue #495, opened 2021-08-18
Respiratory Volume per Time (RVT)


DominiqueMakowski:
Used to detect atypical breathing events, useful in the context of noise removal.

Methods:

- [x] [Birn et al., 2006](http://mriquestions.com/uploads/3/4/5/7/34572113/birn_separating_respirator.pdf)
- [x] [Harrison et al., 2021](https://www.sciencedirect.com/science/article/pii/S1053811921000641)

Resources & Implementations:

- https://github.com/physiopy/phys2denoise

shanelindsay:
RVT would be good to have. 

Just to follow up on those links from Dominique:

The Harrison et al. Hilbert method matlab code is here:
https://github.com/translationalneuromodeling/tapas/blob/master/PhysIO/code/model/tapas_physio_rvt_hilbert.m

More straightforward to integrate would be the python code available here:
https://github.com/physiopy/phys2denoise/blob/master/phys2denoise/metrics/chest_belt.py

Which implements Envelope and RV (respiratory variability).

For noise removal, something like this would also be nice:
https://peakdet.readthedocs.io/en/latest/user_guide/editing.html

PR #695, opened 2022-08-21
[Feature] Added Birn and Power RVT method.

rostro36:
Sorry for the long PR.

# Description
This PR aims at adding more RSP RVT methods as discussed in #495 . I did get the code from [Power's website](https://www.jonathanpower.net/2019-respiratory-measures.html) adjacent to his [paper](https://www.sciencedirect.com/science/article/pii/S1053811919308250?dgcid=author) and translated it from Matlab to Python. I have to check the paper still.

But I think this should add enough methods to #495 and close it.

While implementing I rather used too many parameters, feel free to cut them down. Also I did add the used methods to things like rsp_clean and rsp_peaks. If you would like to have them only in the rvt methods, let me know to change it. The noise I added to the hampel sample maybe could also be added to the nk signal_distort class (or implement it with the already given distortion).

Generally, let me know if I should change things apart from checking the Power paper. I should have time at the latest in the second half of next week. 

# Proposed Changes
## Main
The main addition are the Birn and the Power's method in neurokit2/rsp/rsp_rvt.py.
## Pre-processing
Some special pre-processing is used that uses a different cleaning method and a different peakfinding method, which are also implemented there.
- neurokit2/rsp/rsp_clean.py
- neurokit2/rsp/rsp_findpeaks.py
- neurokit2/rsp/rsp_peaks.py
## Testing
Updated the tests of things that I used:
- tests/tests_rsp.py
- tests/tests_stats.py
## Showcase
You can see some results in rvt_showcase.ipynb. The results seem plausible to me, but I have to read the papers again. The missing signal at the start of Birn and Power come from not having a peak and a trough yet.
# Checklist


DominiqueMakowski:
The plot methods for the power and birn methods don't really work or is it me?

rostro36:
> I streamlined a bit and removed some options, such as the MAD threshold since I don't see a reason why anybody would want to change it :)

Thank you for that :)
> However, the cleaning method is a bit weird, as it can create outliers (see on the right):
> 
> image
> 
> Is it expected?

This can happen if the y-axis difference is too big. If you look at 1200, there we can see the replacement is constant median. 

It is a good sign that only the most extreme values are replaced, not something in the middle. Your threshold of 1 in the sample is quite low, lower than the standard of 3. With the change it looks good, no outliers are found as can be seen in my showcase notebook.

The hampel filter is more optimised for a mostly clean signal with a few signal points with e.g. a 0/NaN reading, like salt and pepper noise in images and what we try to test in the last test.

```python
hampel_sampling_rate=1000
hampel_sample = nk.rsp_simulate(
    duration=duration, sampling_rate=hampel_sampling_rate, respiratory_rate=15, noise=0.1, random_state=42
)
# Random numbers
distort_locations = random.sample(range(len(hampel_sample)), 20)
distorted_sample = copy.copy(hampel_sample)
distorted_sample[distort_locations] = 100
assert np.allclose(
    nk.rsp_clean(distorted_sample, sampling_rate=hampel_sampling_rate, method="hampel", window_length=1),
    hampel_sample,
    atol=1,
)
```

> The plot methods for the power and birn methods don't really work or is it me?

They work for me in the showcase notebook.

DominiqueMakowski:
Sounds good, I'll merge then, thanks a lot for these great improvements!

DominiqueMakowski:
woops it seems i broke it by removing the showcase notebook 🤔

rostro36:
The error came from an ecg test, that we have not touched directly here. But we touched the random module, so I added random_states, which fixed the problem.

DominiqueMakowski:
Awesome, thanks @rostro36

PR #698, opened 2022-09-02
[Docs] Fix documentation of various HRV features

danibene:
# Description

~~Modify the `hrv_time()` function such that the documentation for `MedianNN` reflects what is actually computed and implement the previous description of MedianNN: The median of the absolute values of the successive differences between RR intervals.~~
I corrected some errors in the documentation of `hrv_time()`, `hrv_frequency()`, `hrv_nonlinear()`, `entropy_shannon()`, `entropy_multiscale()`, `fractal_correlation()`, and `fractal_dfa()`.

# Proposed Changes

I changed the `hrv_time()` function as follows:
1. Changed documentation of `MedianNN` and `MCVNN`

~~2. Added `MedianSD` (wasn't sure about the name) as a feature: The median of the absolute values of the successive differences between RR intervals.~~
I also made small changes to the documentation of `hrv_frequency()`, `hrv_nonlinear()`, `entropy_shannon()`, `entropy_multiscale()`, `fractal_correlation()`, and `fractal_dfa()`.

# Checklist
- [] I ran the [CODE CHECKS](https://github.com/neuropsychology/NeuroKit/blob/master/.github/CONTRIBUTING.rst#run-code-checks) on the files I added or modified and fixed the errors.
`astroid.exceptions.TooManyLevelsError: Relative import with too many levels (2) for module 'hrv_time'` I don't think I caused this

### DIFF
--- a/neurokit2/complexity/entropy_multiscale.py
+++ b/neurokit2/complexity/entropy_multiscale.py
@@ -31,7 +31,7 @@ def entropy_multiscale(
     One of the limitation of :func:`SampEn <entropy_sample>` is that it characterizes
     complexity strictly on the time scale defined by the sampling procedure (via the ``delay``
     argument). To address this, Costa et al. (2002) proposed the multiscale entropy (MSEn),
-    which compute sample entropies at multiple scales.
+    which computes sample entropies at multiple scales.
 
     The conventional MSEn algorithm consists of two steps:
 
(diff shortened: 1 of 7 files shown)


DominiqueMakowski:
Is MedianSD described and used in other studies? Wondering why it would be a good addition

danibene:
@DominiqueMakowski 

> Is MedianSD described and used in other studies? Wondering why it would be a good addition

No, not that I know of, but I thought it might have been intentionally described since it seems to me like a good robust alternative for RMSSD. I can revert that change though such that only the description of MedianNN is corrected, if you'd prefer

DominiqueMakowski:
Yeah let's do that. I mean I agree that this index could be a good index, but on the other hand there is such a proliferation of HRV indices already (a pet peeve of mine ^^) that adding new ones without a preliminary validation would do more harm than good to the field

danibene:
@DominiqueMakowski 

> Yeah let's do that. I mean I agree that this index could be a good index, but on the other hand there is such a proliferation of HRV indices already (a pet peeve of mine ^^) that adding new ones without a preliminary validation would do more harm than good to the field

That makes a lot of sense! I've now reverted that commit & added a couple other small fixes to mistakes I noticed while looking at the HRV docs.

DominiqueMakowski:
Wow, these minor but super important fixes are awesome 😍 thanks a lot

Issue #665, opened 2022-06-22
Add preprocessing report feature for quality control


DominiqueMakowski:
I thought there was already an issue for this but couldnt find it.

That's a rather long-term goal. The idea is to add an option; `report=True` or `report="myreport.pdf"`, to processing routines. This would be useful for quality control to check how the preprocessing performed on each individual or sample.

These would contain:

- Figures with: raw signal, filtered signal, peaks detection etc (for visual assessment)
  - We need to find a way to deal with super long signals, do we display it all? do we display a bit to keep the scale reasonable?
- A table containing the mean, SD of for instance heart rate, average signal quality etc. 
  - This table could also be (only) printed in the console when `report=True`
- Specific references of the methods used in the preprocessing (changing depending on the pipeline used)
- Specific text like "The raw ECG signal (sampled at `1000` Hw) was filtered with a bandpass filter ([`0.1, 2`], butterworth 5th order). The peak detection was carried using `authors (2010)` method."

danibene:
> We need to find a way to deal with super long signals, do we display it all? do we display a bit to keep the scale reasonable?

@DominiqueMakowski if you're fine with HTML rather than PDF output you could embed an interactive plot? E.g. with plotly? Alternatively you could show a static plot with the first N seconds (e.g. 0 - 30) and then have a plot for each window of N seconds afterwards (30 - 60, 60 - 90...), maybe in an appendix

danibene:
@DominiqueMakowski 
> We'd need to start with something small that has only a handful of methods / figures

How about we start with `ppg_process()`, with raw/cleaned/peaks + heart rate plots (essentially this but interactive https://neuropsychology.github.io/NeuroKit/_images/p_ppg_process1.png), and then try to expand to `ecg_process()`? 

(I'm biased but) I think even just a tool to check the peak detection for heartbeat signals would be useful

DominiqueMakowski:
sounds good!

danibene:
Is this along the lines of what you had in mind? https://danibene.github.io/neurokit2-ppg-report-example/myreport.html

DominiqueMakowski:
Yeah, that looks awesome! perhaps we should create functions called `ppg_report(method_cleaning="something", method_peak="something")` in which we create the corresponding text programmatically (`if method_cleaning=="this": "Data cleaning was performed using ..."`), so that we can basically call this function and put its output inside the HTML 

(this way, users can also get this info in the console / in a re-useable programmatical form)

[...]

PR #707, opened 2022-09-13
[Feature] Report describing methods in PPG process

danibene:
See https://github.com/neuropsychology/NeuroKit/issues/665

# Description

This PR aims to add the option to output an HTML report describing the methods used to process a PPG signal.

# Proposed Changes

I changed the `ppg_process()` function so that the user can optionally output a report by specifying the name of the report file with the argument `report`. 

(I'm not properly trained in software development best practices so if you think the code could be written in a clearer way, feel free to make suggestions!)

# Checklist


DominiqueMakowski:
(I'll get to this asap, gotta finish some work first :)

danibene:
> Finally found time to give this a look, and I must say it's awesome 😍

🎉 

> I renamed a bunch of stuff (I tend to over-rename things in general whenever I am reviewing code) and reoganized; mostly I separated ppg_report() creation per se from the ppg_methods() sanitation, put the report content wrangling stuff into `misc/`. 

Awesome, honestly this is a lot nicer than me working alone overthinking my own names/organization...

> Before we start adapting it to the rest of the signals, let's make this one top notch, I think we need to:
> 
>     * Somehow test all the combinations to make sure there's no blindspot (not sure how to best test that though)

Were you thinking automatic tests that just confirm that a file is generated without errors? Or more that we generate all examples and confirm that they work as expected?

>     * Add some style to the HTML file
>       
>       * References could be bullet points
>       * Add a title to the report ("NeuroKit Processing Report" (?))
>       * Add some style. This is beyond my abilities though, the best I managed to do recently is to add some CSS in the [preamble](https://github.com/RealityBending/IllusionGameValidation/blob/main/index.html#L3-L27) of an HTML file (which renders like [this](https://realitybending.github.io/IllusionGameValidation/)). We shouldn't spend too much time on style for now though, and keep it miminal, but simply changing the font, having the sections with a background color could already nicely separate the sections
Newest version is now here: https://danibene.github.io/neurokit2-ppg-report-example/myreport.html

>       * Print some of the text in the console as well

This is what the corresponding text now looks like:

```
The raw signal, sampled at 500 Hz, was directly used for peak detection without preprocessing.

The peak detection was carried out using the method described in Elgendi et al. (2013).

Elgendi M, Norton I, Brearley M, Abbott D, Schuurmans D (2013) Systolic Peak Detection in Acceleration Photoplethysmograms Measured from Emergency Responders in Tropical Conditions. PLoS ONE 8(10): e76585. doi:10.1371/journal.pone.0076585.

|   PPG_Rate_Mean |   PPG_Rate_SD |
|----------------:|--------------:|
|         66.9511 |       19.2591 |
```

DominiqueMakowski:
> rather than hard coding them?

Well it would still be hardcoded wouldn't it 😁 The problem of having non .py files in the packages it's that then we need to ship additional files etc. and it complicates a bit the distribution. It's feasible, but if we can keep python files it makes things easier. That being said, we could easily store the same info that you linked in a "hard-coded" python dict instead of a csv/json, there is not much difference (people could easily the contents of such file) afaik?

Though then in general we should also keep in mind that we cannot cater to everyone's need and avoid creating a maintenance beast for us: the report should be geared as a convenient way to get the info in a reproducible manner, but then if people want more control and customization and all, they might as well re-create it themselves to be how they want

> Were you thinking automatic tests that just confirm that a file is generated without errors? Or more that we generate all examples and confirm that they work as expected?

I guess both, probably the first (that it doesn't error) is more important, then if there are some issues in the report's content I'm sure users will report them in issues.

danibene:
> The problem of having non .py files in the packages it's that then we need to ship additional files etc. and it complicates a bit the distribution. It's feasible, but if we can keep python files it makes things easier.

That makes sense

> I guess both, probably the first (that it doesn't error) is more important, then if there are some issues in the report's content I'm sure users will report them in issues.

I added a test, let me know if you think we should change it

DominiqueMakowski:
Looks very good to me, feel free to merge!

Issue #684, opened 2022-07-22
RRI & RRI_Time as input to hrv()


danibene:
**Describe the solution you'd like**
I would like the HRV functions to be able to accept interbeat intervals and their corresponding timestamps as input rather than the peak samples. This seems to be a recurring issue (e.g. https://github.com/neuropsychology/NeuroKit/issues/682, https://github.com/neuropsychology/NeuroKit/issues/667). I also think it would add more flexibility in terms of how to preprocess the interbeat intervals before HRV analysis (outlier removal & treatment: https://github.com/neuropsychology/NeuroKit/issues/541, interpolating & detrending: https://github.com/neuropsychology/NeuroKit/pull/680).

**How could we do it?**
I did this for a project where I had missing data & therefore the timestamps of the interbeat intervals were needed to properly interpolate in hrv_frequency(). I also made some other changes to the HRV analysis functions to account for the missing data (but I think the input format is probably better to address first on its own). Here is the current state of that branch:
https://github.com/danibene/NeuroKit/blob/feature/adjust_for_missing_intervals/neurokit2/hrv/hrv_time.py#L148-L170

There, I added extra arguments `rri_time` and `data_format` but in retrospect, it would probably be cleaner if there were no extra input arguments and if the input type could just be automatically detected from the keys of a dictionary/columns of a dataframe ("RRI", "RRI_Time")?

Let me know if you are interested in this :)

DominiqueMakowski:
Seems I totally missed this (guess the stale bot *can* be useful), what's the status of this?

DominiqueMakowski:
Agree to check the peaks argument, best is to keep the number of arguments tight when possible

Issue #541, opened 2021-09-08
Workflows for outlier removal


shanelindsay:
I am working with respiration data right now, but I think is a more general question. 

I am unclear on the workflows for outlier detection and removal. 

One way to remove outliers (e.g. movement artefacts with spikes in data much outside normal range) is manual inspection and either removal of periods of bad data or adjustment of events (e.g. remove/change detected peaks). Some tools use a gui ([e.g. see eeglab](https://eeglab.org/tutorials/06_RejectArtifacts/Scrolling_data.html) to facilitate this which I imagine would be on a "nice to have" list for future neurokit. 

Alternatively you could use some code for outlier detection/removal.

Either way, I am a bit unclear on what is expected in the neurokit way of doing things. Arrays aren't indexed by time (instead by sample) so removal of periods of bad data would mess up timing. Setting bad data to NA would also seem like it would mess up a lot of algorithms and get undesired results (also gets tricky with interval/epochs). 

Other tools have more complicated data structures that incorporate removed data e.g. eeglab "EEGLAB adjust the EEG.event structure fields and insert boundary events where data has been rejected, with a duration field holding the duration of the data portion that was rejected".

But for neurokit, I was curious on the recommended current workflows for dealing with outliers and any future plans.

DominiqueMakowski:
Thanks again Shane for sharing some ideas. I think there are essentially two different types of "bad data":

One would be "transcient artefacts", resulting in sudden spikes that could be mistakenly detected as peaks. This, primarily, would be addressed by filtering, [peak fixing](https://github.com/neuropsychology/NeuroKit/blob/master/neurokit2/signal/signal_fixpeaks.py), etc. It's true that GUIs can be helpful in this case to manually select / deselect artifacts, but manual correction has not really been the direction of the package. I'm not exactly sure how to give to users more control to manually adjust the detected peaks... 

The second class of bad data would be indeed bad periods of data, i.e., where data is simply too noisy for some time due to disconnection, movement, or whatnot. For this we don't really have any tools in NK, and I can foresee that it's quite a hassle to deal with as you mentioned: if we turn to NaNs all the values of the periods, then we would need to run the processing on all the periods of good data separately, which then leads to other issues. 
I think so far we've swept this issue under the rug, but it is true that periods of bad data are fairly common. The list of things to clarify API-wise would be:

[...]

PR #710, opened 2022-09-21
[Feature] R-R intervals and associated timestamps as input to HRV functions

danibene:
# Description

This PR aims at adding the ability to directly pass R-R intervals to the HRV functions. See also https://github.com/neuropsychology/NeuroKit/issues/684

# Proposed Changes

Instead of separately sanitizing the input and converting to R-R intervals, I added a function `_hrv_format_input()` that can accept a dictionary with the keys `RRI` and `RRI_Time` and always returns the same format (either peaks or intervals depending on the HRV function) regardless of whether peaks or intervals are passed.

# Checklist


DominiqueMakowski:
image

I think the docstrings of intervals_to_peaks failed 🤔

danibene:
> image
> 
> I think the docstrings of intervals_to_peaks failed 🤔

I think I fixed it now! 

Since there's a bit of repetition in the way the intervals are cleaned across functions (`_hrv_sanitize_rri()`, `intervals_to_peaks()`), thoughts on refactoring such that `intervals` is a separate subpackage containing e.g. `intervals_to_peaks()`, `find_successive_intervals()`, `intervals_sanitize()`, and `intervals_preprocess()`? Or should I just leave as is?

DominiqueMakowski:
mmh a subpackage might be overkill, but we could put all these functions in separate files within the hrv module, since HRV is the one that predominantly deals with intervals?

- intervals_to_peaks()
- find_successive_intervals(): could be renamed `intervals_successive()` for consistency?
- intervals_sanitize()
- intervals_preprocess()

Or perhaps just putting these in separate files in the `misc`

DominiqueMakowski:
What about merging intervals_successive with intervals_sanitize()? Or do you think there would be some usecase for intervals_successive() outside of intervals sanitization?

DominiqueMakowski:
If I understand you use it through `_hrv_format_input` which uses it via sanitize in `_hrv_get_rri()` anyway, so we could make it an internal function and put it in hrv_utils no? 

(my thinking is that if a function is mostly used by us or rather than by users in their analysis, it can be good to make it an internal to prevent [feature creep](https://en.wikipedia.org/wiki/Feature_creep) and avoid bloating the documentation - that's why I was asking for potential other use cases that users my have with this)

danibene:
> If I understand you use it through `_hrv_format_input` which uses it via sanitize in `_hrv_get_rri()` anyway, so we could make it an internal function and put it in hrv_utils no?
> 
> (my thinking is that if a function is mostly used by us or rather than by users in their analysis, it can be good to make it an internal to prevent [feature creep](https://en.wikipedia.org/wiki/Feature_creep) and avoid bloating the documentation - that's why I was asking for potential other use cases that users my have with this)

Yes, but I also plan to use `intervals_successive()` within the `hrv_time()` and `hrv_nonlinear()` functions in the PR addressing https://github.com/neuropsychology/NeuroKit/issues/541 :

My current implementation is here: https://github.com/danibene/NeuroKit/blob/feature/adjust_for_missing_intervals/neurokit2/hrv/hrv_time.py#L168-L173

I can still make it an internal function though - should I move it to `_intervals_successive()` in `hrv_utils`?

DominiqueMakowski:
okay cool

> I can still make it an internal function though - should I move it to _intervals_successive() in hrv_utils?

let's do that for now (we can leave all the examples and descriptions), and down the road we can always make it user-facing if need be

PR #712, opened 2022-09-27
[Feature] use lombscargle PSD method without interpolation

danibene:
# Description
This PR aims to add the ability to use the lombscargle PSD method without interpolating the signal, i.e., so that (possibly unevenly spaced) timestamps can be directly used. See also  https://github.com/neuropsychology/NeuroKit/issues/671#issuecomment-1192153962

# Proposed Changes

1. I changed the `hrv_frequency()` function so that the timestamps `t` are passed as an argument to `signal_psd()` if the `interpolation_rate` argument is set to `None`
2. I added the timestamps `t` as an argument to `signal_psd()` so they can be passed to `_signal_psd_lomb()`
3. I added functions to check whether the timestamps were already interpolated to be evenly spaced (`_intervals_time_uniform()`) and if so, determine their sampling rate (`_intervals_time_to_sampling_rate()`), so that the user can pass interpolated R-R intervals to `hrv_frequency()` and these are still treated as interpolated by `signal_psd()`.

# Checklist

### DIFF
--- a/neurokit2/hrv/intervals_process.py
+++ b/neurokit2/hrv/intervals_process.py
@@ -17,7 +17,7 @@ def intervals_process(
     intervals_time=None,
     interpolate=False,
     interpolation_rate=100,
-    detrend_method=None,
+    detrend=None,
     **kwargs
 ):
     """**Interval preprocessing**
@@ -33,7 +33,7 @@ def intervals_process(
     interpolation_rate : int, optional
         Sampling rate (Hz) of the interpolated interbeat intervals. Should be at least twice as
         high as the highest frequency in vhf. By default 100. To replicate Kubios defaults, set to 4.
-    detrend_method : str
+    detrend : str
         Can be one of ``"polynomial"`` (traditional detrending of a given order) or
         ``"tarvainen2002"`` to use the smoothness priors approach described by Tarvainen (2002)
         (mostly used in HRV analyses as a lowpass filter to remove complex trends), ``"loess"`` for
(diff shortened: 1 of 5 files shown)


neurokit2/hrv/intervals_process.py
```diff
@@ -102,6 +108,12 @@ def intervals_process(
         )
 
         intervals = signal_interpolate(intervals_time, intervals, x_new=x_new, **kwargs)
+    else:
+        # check if intervals appear to be already interpolated
+        if _intervals_time_uniform(intervals_time):
+            # get sampling rate used for interpolation
+            interpolation_rate = _intervals_time_to_sampling_rate(intervals_time)
+
     if detrend_method is not None:
```
DominiqueMakowski, line 117: perhaps i would have this argument named just `detrend = None` (for no detrending) or a string with a specific method name?

neurokit2/hrv/intervals_process.py
```diff
@@ -5,15 +5,19 @@
...
+)
 
 
 def intervals_process(
     intervals,
     intervals_time=None,
     interpolate=False,
     interpolation_rate=100,
-    detrend_method=None,
+    detrend=None,
```
DominiqueMakowski, line 20: should we detrend by default tho?

danibene: No I wouldn't, since it would be a major change from the current implementation and also it shouldn't be done for all hrv features (e.g. MeanNN, MaxNN)

DominiqueMakowski:
Looks good to me, feel free to merge when you want

RichardBJ:
Ah, I was on an old version of this and about to suggest this very feature. The whole point of a Lomb surely is exactly this, unevenly spaced data! Going to try it later :-) Thanks!!

RichardBJ:
Ah, yikes not working for me.  IS this the correct place to discuss this, or should I open an issue? Thanks and so sorry!

danibene:
> Ah, yikes not working for me.  IS this the correct place to discuss this, or should I open an issue? Thanks and so sorry!

@RichardBJ oh no! Don't be sorry, thanks for calling our attention to it :) opening an issue with a reproducible example would be great if possible

Issue #558, opened 2021-10-12
ECG peak detection robust to signal inversion


raphaelvallat:
**Question and context**

Hi here! 

Great work on the package 👏 👏  

I am analyzing a large database of ECG recordings (n=2000+) and I suspect that for some of those (but not all), the ECG data is inverted. I do not have the time to do a manual check of all the files. My question is therefore whether there is a preferred method that I should use in `ecg_clean` and `ecg_peaks` that will be robust to such polarity inversion. I have been using the default method ("neurokit") and it seems to be working quite well, but maybe there's a better alternative.

PS: I think this is related to https://github.com/neuropsychology/NeuroKit/issues/368.

Thanks for your help!
Raphael

DominiqueMakowski:
Did you have any luck with that in the end?

raphaelvallat:
Nope, ended up working on other projects and never had the time to finish this. Feel free to close the issue!

DominiqueMakowski:
FYI @raphaelvallat, @danibene implemented a new `ecg_invert()` function that detects whether inversion is needed (and inverts it if need be). Closing this then :)

PR #716, opened 2022-09-29
[Feature] ECG inversion

danibene:
# Description

This PR aims at adding the ability to identify and correct ECG signal inversion. See also https://github.com/neuropsychology/NeuroKit/issues/558

# Proposed Changes

I added the `ecg_invert()` function so that a user could optionally correct for signal inversion in the raw ECG data before using the other processing functions (cleaning, peak detection).

# Checklist

### DIFF
--- a/docs/functions/ecg.rst
+++ b/docs/functions/ecg.rst
@@ -71,7 +71,9 @@ Miscellaneous
 """"""""""""""""""""""""
 .. autofunction:: neurokit2.ecg.ecg_findpeaks
 
-
+*ecg_invert()*
+""""""""""""""""""""""""
+.. autofunction:: neurokit2.ecg.ecg_invert
 
 *Any function appearing below this point is not explicitly part of the documentation and should be added. Please open an issue if there is one.*
 
--- a/docs/functions/ecg.rst
+++ b/docs/functions/ecg.rst
@@ -79,4 +79,4 @@ Miscellaneous
 
 .. automodule:: neurokit2.ecg
    :members:
-   :exclude-members: ecg_simulate, ecg_analyze, ecg_process, ecg_plot, ecg_clean, ecg_quality, ecg_rsp, ecg_rsp, ecg_peaks, ecg_delineate, ecg_phase, ecg_rate, ecg_segment, ecg_eventrelated, ecg_intervalrelated, ecg_findpeaks
+   :exclude-members: ecg_simulate, ecg_analyze, ecg_process, ecg_plot, ecg_clean, ecg_quality, ecg_rsp, ecg_rsp, ecg_peaks, ecg_delineate, ecg_phase, ecg_rate, ecg_segment, ecg_eventrelated, ecg_intervalrelated, ecg_findpeaks, ecg_invert
--- a/neurokit2/ecg/ecg_invert.py
+++ b/neurokit2/ecg/ecg_invert.py
@@ -55,10 +55,13 @@ def ecg_invert(ecg_signal, sampling_rate=1000, check_inverted=True):
         return inverted_ecg_signal
 
 
-def _ecg_inverted(ecg_signal, sampling_rate=1000):
+def _ecg_inverted(ecg_signal, sampling_rate=1000, window_time=2.0):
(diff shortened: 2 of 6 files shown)


docs/functions/ecg.rst
```diff
@@ -71,7 +71,9 @@ Miscellaneous
 """"""""""""""""""""""""
 .. autofunction:: neurokit2.ecg.ecg_findpeaks
 
-
+*ecg_invert()*
+""""""""""""""""""""""""
+.. autofunction:: neurokit2.ecg.ecg_invert
 
 *Any function appearing below this point is not explicitly part of the documentation and should be added. Please open an issue if there is one.*
```
DominiqueMakowski, line 79: we should also add any new function to the exclude-members below, basically the section below documents all the functions from - in this case - the ecg module with the exceptions of the ones I specified: this was used as a check for me to make sure I didn't forget any functions but if we don't add new functions to the exclusion it will be documented two times

neurokit2/ecg/ecg_invert.py
```diff
@@ -0,0 +1,72 @@
+import numpy as np
+
+from .ecg_clean import ecg_clean
+
+
+def ecg_invert(ecg_signal, sampling_rate=1000, check_inverted=True):
+    """**ECG signal inversion**
+
+    Checks whether an ECG signal is inverted, and if so, corrects for this inversion.
```
DominiqueMakowski, line 9: maybe specify a bit how the check is done

neurokit2/ecg/ecg_invert.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def _ecg_inverted(ecg_signal, sampling_rate=1000):
+    """Checks whether an ECG signal is inverted."""
+    ecg_cleaned = ecg_clean(ecg_signal, sampling_rate=sampling_rate)
+    med_max = np.nanmedian(_roll_func(ecg_cleaned, window=1 * sampling_rate, func=_orig_max_squared))
+    return med_max < np.nanmean(ecg_cleaned)
+
+
+def _roll_func(x, window, func, func_args={}):
```
DominiqueMakowski, line 64: isn't it more efficient to use the built-in pandas' rolling method? Efficiency is not critical here, but just wondering

neurokit2/ecg/ecg_invert.py
```diff
@@ -0,0 +1,72 @@
...
+    ecg_signal : Union[list, np.array, pd.Series]
+        The raw ECG channel.
+    sampling_rate : int
+        The sampling frequency of ``ecg_signal`` (in Hz, i.e., samples/second). Defaults to 1000.
+    check_inverted : bool, optional
+        Whether to check whether the signal is inverted before inverting it. The default is True.
+        If False, always returns the inverted input signal regardless of whether the input was inverted.
+    Returns
+    -------
+    array
```
DominiqueMakowski, line 22: perhaps we could consistently return a tuple, 1) the returning signal and 2) a bool of whether inversion was performed or not (i.e., whether it was detected as needed).

Then we could replace the check_inverted arg by `force=False` to force the inversion, in which case the return would be `(inverted_ecg, True)`

neurokit2/ecg/ecg_invert.py
danibene: Seems like it's better to use numpy, so I re-implemented that: https://gist.github.com/danibene/ada261b492b36daeeb1acdb34be44032

DominiqueMakowski:
I streamlined the example a bit and added a `show` option. Feel free to adjust/modify!

(Also, note that I added some 

```
@suppress
plt.close()
```

After the plots: the reason for this is that basically, when sphinxruns the documentation and runs ipython on the example chuncks, it will create matplotlib figures. But it doesn't close them automatically, meaning that after 20 opened images it will error and say that there are too many images concurrently opened (and the documentation build will fail). So we need to close the plots that we create in the docstrings manually)

danibene:
@DominiqueMakowski this all looks good to me, I'll go ahead and merge!

Issue #692, opened 2022-08-10
[Feature] Simplify code checks for contributors


rostro36:
**Describe the solution you'd like**
Currently, to run the code checks, there are five lines that need to be run per file/directory. This is a bit clunky, especially for many files and it was also mentioned before in #321 .

This should not make a difference to the whole actions in GitHub or installation of the package for users.

These scripts are what I was talking about:

> isort myfile.py -l 120  --balanced --multi-line 3 --lines-between-types 1 --lines-after-imports 2 --trailing-comma
> black myfile.py --line-length 120
> docformatter myfile.py --wrap-summaries 120 --wrap-descriptions 113 --blank --in-place
> 
> flake8 myfile.py --max-line-length=127 --max-complexity=10 --ignore E303,C901,E203,W503
> pylint myfile.py --max-line-length=127 --load-plugins=pylint.extensions.docparams --load-plugins=pylint.extensions.docstyle --variable-naming-style=any --argument-naming-style=any --reports=n --suggestion-mode=y --disable=E303 --disable=R0913 --disable=R0801 --disable=C0114 --disable=E203 --disable=E0401 --disable=W9006 --disable=C0330 --disable=R0914 --disable=R0912 --disable=R0915 --disable=W0102 --disable=W0511 --disable=C1801 --disable=C0111 --disable=R1705 --disable=R1720 --disable=C0301 --disable=C0415 --disable=C0103 --disable=C0302 --disable=R1716 --disable=W0632 --disable=E1136 --extension-pkg-whitelist=numpy

**How could we do it?**
Doing a little bit of research, I found two kinds of solutions:
- **hook e.g. [pre-commit](https://pre-commit.com/) or [git hook](https://githooks.com/)**
     - ✔️ Pro
           - Can't forget as it is automatic
           - Automatically only looks at differences
           - Industry standard
     - ❌ Contra
          - Gets executed every time, even for the smallest commits and can take minutes of waiting
          - Denies action (e.g. commit) if there is an error (I assume this can be disabled). This is especially problematic if the code already has legacy errors not from the new commit. This might frustrate contributors not used to it.
          - It is rather a side effect, that may not be too easy to understand.
- **script e.g. bash** 
    - ✔️ Pro
         - Easier to understand for beginners 
         - Still allows the action to go through and can be fixed later until found during merge checks.
    - ❌ Contra 
         - Easier to forget and also contributor might not check for all files.
         - Surely not elegant.
[...]

PR #720, opened 2022-10-03
[Feature] adjust HRV functions for missing intervals

danibene:
# Description

This PR aims at adding the ability to perform HRV analysis in the case of missing interbeat intervals. See also https://github.com/neuropsychology/NeuroKit/issues/541

# Proposed Changes

I changed the `hrv_time()` and `hrv_nonlinear()` functions so that for some features, differences between intervals that are not actually successive are removed and the timestamps of the intervals are taken into account. 

# Checklist
(Maybe this checklist should be updated to remove the code checks & News.rst points? https://github.com/neuropsychology/NeuroKit/issues/692#issuecomment-1211437961)

### DIFF
--- a/tests/tests_hrv.py
+++ b/tests/tests_hrv.py
@@ -84,11 +84,11 @@ def test_hrv_detrended_rri(detrend):
     rri = np.diff(peaks).astype(float)
     rri_time = peaks[1:] / 1000
 
-    rri_processed, _ = nk.intervals_process(
+    rri_processed, rri_processed_time, _ = nk.intervals_process(
         rri, intervals_time=rri_time, interpolate=False, interpolation_rate=None, detrend=detrend
     )
 
-    ecg_hrv = nk.hrv({"RRI": rri_processed, "RRI_Time": rri_time})
+    ecg_hrv = nk.hrv({"RRI": rri_processed, "RRI_Time": rri_processed_time})
 
     assert np.isclose(ecg_hrv["HRV_RMSSD"].values[0], np.sqrt(np.mean(np.square(np.diff(rri_processed)))), atol=0.1)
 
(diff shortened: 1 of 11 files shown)


DominiqueMakowski:
just a naive question, in general, rather than having `rri_missing` arg in every subfunction, wouldn't it be easier to just filter out all missing before and pass the valid RRi list?

danibene:
> just a naive question, in general, rather than having `rri_missing` arg in every subfunction, wouldn't it be easier to just filter out all missing before and pass the valid RRi list?

My thinking here was that since only the valid RRI are passed but the `intervals_successive()` function wouldn't work in the case of interpolated/detrended R-R intervals (as the difference between timestamps wouldn't match up with the intervals), it would be better to just do a single check at the beginning to try to detect whether there are missing intervals and then adjust functions accordingly. 

[I also thought about whether there should be some kind of option in the input data to specify how missing data should be detected, to be extra sure that interpolated/detrended R-R intervals are not accidentally detected as containing missing data?](https://github.com/neuropsychology/NeuroKit/issues/541#issuecomment-1262537481) Let me know if you think that should be added.

DominiqueMakowski:
Mmh that's a tough one... I understand your point but somehow it seems like a lot of code and logic for a few potential missing. 
Just to confirm, these missing values are the result of the detection of non-consecutive intervals right? It's not something passed by the user?

danibene:
@DominiqueMakowski yes for functions like pNN50, RMSSD, SD1, differences that are detected as non-consecutive are removed and this is based on the timestamps, not only e.g. NaNs explicitly passed by the user. The reason for this was because when I had missing data, the signal was too noisy for peak detection & so I didn't know how many missing intervals there actually were (/how many NaNs should be inserted), just the timestamps of the intervals when the signal stopped being as noisy again.

tests/tests_hrv.py
```diff
@@ -102,14 +102,21 @@ def test_hrv_interpolated_rri(interpolation_rate):
...
+
     if interpolation_rate=="from_mean_rri":
         interpolation_rate = 1000/np.mean(rri)
+
     rri_processed, _ = nk.intervals_process(
         rri, intervals_time=rri_time, interpolate=True, interpolation_rate=interpolation_rate
     )
 
-    ecg_hrv = nk.hrv({"RRI": rri_processed, "RRI_Time": rri_time})
+    rri_processed_time = np.arange(
```
danibene, line 113: I wonder if `intervals_process()` should return the new timestamps so that they don't have to be recomputed outside of the function. Maybe only if `intervals_time` is passed?

DominiqueMakowski: I'd say you know best :)

danibene:
Since the tests are passing even with detrended/interpolated input I think it's fine to merge now

DominiqueMakowski:
Alright feel free to merge then :)

Issue #728, opened 2022-10-11
Defaults for detrending with locreg method


danibene:
I noticed that the `locreg` method of `signal_detrend()` has the default `stepsize=0.02`, but if `stepsize` is less than 1, an error will be raised:

https://github.com/neuropsychology/NeuroKit/blob/5cd089e9d82e6f7fb98d2028e69840aee08f8ab3/neurokit2/signal/signal_detrend.py#L176-L198

Looking at one of the scripts the `locreg` method is based off of, the default step size is also `0.02`, but it seems to be in seconds, not samples:
https://github.com/sappelhoff/pyprep/blob/master/pyprep/removeTrend.py#L92-L103

Some ways this could be addressed:
- Add `sampling_rate` as an optional argument for `signal_detrend()` 
- Change the default to match a sampling rate of 1000 (which seems to generally be the default in NeuroKit), i.e. 0.02 seconds would be 20 samples

PR #730, opened 2022-10-12
[Fix] defaults detrend locreg

danibene:
# Description

See https://github.com/neuropsychology/NeuroKit/issues/728

# Proposed Changes

I added `sampling_rate` as an argument to the `signal_detrend()` function.

# Checklist


neurokit2/signal/signal_detrend.py
```diff
@@ -44,15 +45,19 @@ def signal_detrend(
...
-        desired low frequency band to remove multiplied by the sampling rate
-        (for instance, ``1.5*1000`` will remove frequencies below 1.5Hz for a signal sampled at
-        1000Hz).
+        desired low frequency band to remove multiplied by the sampling rate if sampling_rate is not
+        provided to the function directly (for instance, ``1.5*1000`` will remove frequencies below 
+        1.5Hz for a signal sampled at 1000Hz).
     stepsize : float
         Only used if ``method`` is ``"locreg"``. Similarly to ``window``, ``stepsize`` should also
-        be multiplied by the sampling rate.
+        be multiplied by the sampling rate if sampling_rate is not provided.
```
DominiqueMakowski, line 53: if ~not~ provided?

neurokit2/signal/signal_detrend.py
```diff
@@ -15,6 +15,7 @@ def signal_detrend(
     window=1.5,
     stepsize=0.02,
     components=[-1],
+    sampling_rate=None,
```
DominiqueMakowski, line 18: we should put it to 1000 by default given that default stepsize is not an integer, 

also we should probably pipe down the sampling rate in the routines where we use it (ecg_process?)

danibene: currently there don't seem to be any functions that call the locreg method of detrending (or have a parameter for specifying the method of detrending), other than intervals_process which I changed

neurokit2/signal/signal_detrend.py
danibene: I just left it out completely now, feel free to change if you think it should be specified

DominiqueMakowski (approved):

Issue #729, opened 2022-10-11
"[N] not in index" exception raised by ecg_delineate


citron:
**Bug Description:**
ECG having length N, delineation stops on an exception with "[N] not in index" as a debug message.

**To Reproduce:**

1. save following code as debug_pyatsa.py in your favorite folder
```
import neurokit2 as nk
import pandas as pd
import numpy as np
import os

HERTZ = 250
filin = "debug.csv"
ecg_signal = pd.read_csv(filin, header=0, names=["ECG"])['ECG']
ecg_cleaned = nk.ecg_clean(ecg_signal, sampling_rate=HERTZ)
_, rpeaks = nk.ecg_peaks(ecg_cleaned, sampling_rate=HERTZ)
_, waves_peak = nk.ecg_delineate(ecg_cleaned, rpeaks, sampling_rate=HERTZ, method="peak", show=False, show_type='peaks')
```
2. save attached ECG sample as debug.csv 
[debug.csv](https://github.com/neuropsychology/NeuroKit/files/9758396/debug.csv)
3. python debug_pyatsa.py

**Expected behaviour:**
No exception should occur as the CSV input file is OK.

**Screenshots**
![Capture d’écran 2022-10-11 184334](https://user-images.githubusercontent.com/161367/195151319-dd5d7e11-a33c-448e-a674-ccce82437988.jpg)
![Capture d’écran 2022-10-11 192032](https://user-images.githubusercontent.com/161367/195158085-bb123e39-0ded-404d-bda5-cb1fa59dc328.jpg)

**System Specifications:**

- OS: Linux (ELF 64bit) 
- Python: 3.10.7 
- NeuroKit2: 0.2.1 
- NumPy: 1.23.3 
- Pandas: 1.5.0 
- SciPy: 1.9.1 
- sklearn: 1.1.2 
- matplotlib: 3.6.1

same problem on Windows :
 - OS: Windows (WindowsPE 64bit) 
- Python: 3.10.7 
- NeuroKit2: 0.2.1 
- NumPy: 1.23.3 
- Pandas: 1.5.0 
- SciPy: 1.9.2 
- sklearn: 1.1.2 
- matplotlib: 3.6.1

DominiqueMakowski:
@Tam-Pham time to revise your python skills 😁

citron:
I do confirm. It works now using the dev branch. Thank you very much.

PR #731, opened 2022-10-13
[Fix] ECG delineator peak index

danibene:
# Description

This PR aims to fix an error in `ecg_delineate()` where invalid indices were returned. See also https://github.com/neuropsychology/NeuroKit/issues/729

# Proposed Changes

I changed the `ecg_delineate()` function so that within `_ecg_delineator_peak()`, all indices of fiducial points are checked to ensure that they are not larger than the number of samples in the ECG signal - 1

# Checklist

- [x] I ran the [CODE CHECKS](https://github.com/neuropsychology/NeuroKit/blob/master/.github/CONTRIBUTING.rst#run-code-checks) on the files I added or modified and fixed the errors. (I only actively corrected sections of the code that I modified here)

### DIFF
--- a/neurokit2/ecg/ecg_delineate.py
+++ b/neurokit2/ecg/ecg_delineate.py
@@ -19,14 +19,7 @@ from .ecg_segment import ecg_segment
 
 
 def ecg_delineate(
-    ecg_cleaned,
-    rpeaks=None,
-    sampling_rate=1000,
-    method="dwt",
-    show=False,
-    show_type="peaks",
-    check=False,
-    **kwargs
+    ecg_cleaned, rpeaks=None, sampling_rate=1000, method="dwt", show=False, show_type="peaks", check=False, **kwargs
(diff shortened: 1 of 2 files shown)


danibene:
Also should we add https://github.com/neuropsychology/NeuroKit/files/9758396/debug.csv to https://github.com/neuropsychology/NeuroKit/tree/master/tests/ecg_data & use the code that led to the error as a test for `ecg_delineate()`? I just tested it by installing my branch:

Pip install: https://gist.github.com/danibene/8b5df2192b5a063d10e59972d77556d6 
Installing my branch: https://gist.github.com/danibene/6cb93f7c2654698b91f0f3a5de2e8166 

but maybe it'd be good to have it as a proper test?

DominiqueMakowski (approved):
Ah nice, thanks for digging into the delineator code, it's a tough one :)

PR #734, opened 2022-10-18
[Docs] window size calculation for locreg detrending

vasileermicioi:
# Description

This PR aims to fix docs for detrending.

# Proposed Changes

Doc comments were updated.

# Checklist

- [x] I ran the [CODE CHECKS](https://github.com/neuropsychology/NeuroKit/blob/master/.github/CONTRIBUTING.rst#run-code-checks) on the files I added or modified and fixed the errors. (not applicable)
- [x] I have added the newly added features to **News.rst** (not applicable)


vasileermicioi:
The bug was in using a window of 0.5 * sample_rate, which corresponds to a frequency of 2Hz, not to a frequency of 0.5Hz as desired.

A few comments to clarify

1) Any average method is a lowpass filter including simple averages, moving averages, Savitzki-Golay, etc.

2) Formula is window=(sample_rate/frequency).
E.g. if the sample rate is 1000Hz and we want to keep frequencies below 8Hz, then the window size will be 1000/8=125 

3) Heart rate information is in the range of 0.5Hz-5Hz, and, in order not to distort the signal, only filters that remove frequencies below 0.5Hz are used (lowpass filters). If the heart rate is known, then the perfect detrend is below the heart rate frequency (heart rate frequency = heart rate / 60).

4) Any lowpass filter can be used as a detrending method, including Butterworth and Chebyshev, and after the bandpass, the signal is already detrended.
For e.g. applying a bandpass filter of 0.5hz-15hz will detrend (frequencies below 0.5Hz are removed) and smooth (frequencies above 15Hz are removed)  the signal

vasileermicioi:
2 examples of signals with the heart rate of 60bpm and 120bpm respectively, filtered with 0.5Hz and 2Hz.
You can see that the 2Hz filter distorted the signal with bpm=60, 
and the 0.5Hz filter on the signal with bpm=120 left didn't remove the trend completely.

60bpm_locreg_2hz
60bpm_locreg_05hz
60bpm_original
120bpm_locreg_2hz
120bpm_locreg_05hz
120bpm_original

vasileermicioi:
Here 2 examples of PPG signals with the heart rate of 60bpm and 120bpm respectively, filtered with 0.5Hz and 2Hz.
You can see that the 2Hz filter distorted the signal with bpm=60,
and the 0.5Hz filter on the signal with bpm=120 left didn't remove the trend completely.

the 60 bpm signal
60bpm_original
60bpm_locreg_05hz
60bpm_locreg_2hz

the 120 bpm signal
120bpm_original
120bpm_locreg_05hz
120bpm_locreg_2hz

danibene:
Thank you for this @vasileermicioi ! I added some comments above. 

> * Heart rate information is in the range of 0.5Hz-5Hz, and, in order not to distort the signal, only filters that remove frequencies below 0.5Hz are used (lowpass filters). If the heart rate is known, then the perfect detrend is below the heart rate frequency (heart rate frequency = heart rate / 60).

Do you happen to have a reference for this? I agree that we would not filter out frequencies above 0.5Hz for interbeat interval detrending, but for preprocessing a raw heartbeat signal like ECG, I've seen higher cutoff frequencies e.g. Pan-Tompkins uses 5 Hz: https://github.com/neuropsychology/NeuroKit/blob/e66a73222de29edeba677f8b33c7179196c3fd97/neurokit2/ecg/ecg_clean.py#L191-L196

No worries if not, but how the appropriate cutoff frequency is determined is something I've wondered myself!

vasileermicioi:
> 
heart rate information is in 0.5Hz-5Hz because the heart rate is in 30-300bpm (multiply frequency by 60 :) ),
but usual high-cut for PPG is higher than 5Hz like 8Hz, 10Hz or 15Hz in order to preserve some details like diastolic peaks,
(https://www.researchgate.net/publication/328256630_Sinus_or_not_a_new_beat_detection_algorithm_based_on_a_pulse_morphology_quality_index_to_extract_normal_sinus_rhythm_beats_from_wrist-worn_photoplethysmography_recordings)

for ECG it is 40Hz, see https://www.medteq.net/article/2017/4/1/ecg-filters

danibene:
Thank you for the references @vasileermicioi !

> heart rate information is in 0.5Hz-5Hz because the heart rate is in 30-300bpm (multiply frequency by 60 :) ), but usual high-cut for PPG is higher than 5Hz like 8Hz, 10Hz or 15Hz in order to preserve some details like diastolic peaks, (https://www.researchgate.net/publication/328256630_Sinus_or_not_a_new_beat_detection_algorithm_based_on_a_pulse_morphology_quality_index_to_extract_normal_sinus_rhythm_beats_from_wrist-worn_photoplethysmography_recordings)
> 
> for ECG it is 40Hz, see https://www.medteq.net/article/2017/4/1/ecg-filters

I meant that the low-cut of Pan-Tompkins, an algorithm used to process ECG signals, is 5 Hz (the high-cut is 15 Hz), so I was curious if you had any sources for digitally filtering out frequencies below 5 Hz being necessarily problematic?

vasileermicioi:
> I meant that the low-cut of Pan-Tompkins, an algorithm used to process ECG signals, is 5 Hz (the high-cut is 15 Hz), so I was curious if you had any sources for digitally filtering out frequencies below 5 Hz being necessarily problematic?

you are right about ECG having a different range for bandpass than PPG, my experience and example are with PPG, and my statement is true for PPG only,
https://www.elgendi.net/papers/QRS_Bands_final.pdf
this article compares ECG bandpass bands and suggests 8Hz-20Hz being optimal,
while for PPG, it is usually 0.5Hz-8Hz

other than that the PR is still correct as 
- docs for the window calculation  for the `locreg` method is computed incorrectly, it should be window = sample_rate / freq, and not window=sample_rate * freq
- for consistency as the other 2 methods (polynomial and standardized) are using 0.5Hz for lowcut, so it should locreg
- using lowcut frequencies above 0.5Hz for PPG will distort the signal (see my example)

PS:
At first, I thought the polynomial is wrong too, but using an additional order for every 2 seconds (in the code is length / 2), has the same effect as `locreg` with window=2*sample_rate (which is the same as window=sample_rate/0.5, just in case :) )

danibene:
> you are right about ECG having a different range for bandpass than PPG, my experience and example are with PPG, and my statement is true for PPG only, https://www.elgendi.net/papers/QRS_Bands_final.pdf this article compares ECG bandpass bands and suggests 8Hz-20Hz being optimal, while for PPG, it is usually 0.5Hz-8Hz

Ah interesting, thank you :)

> other than that the PR is still correct as
> 
>   docs for the window calculation  for the `locreg` method is computed incorrectly, it should be window = sample_rate / freq, and not window=sample_rate * freq
 Were you able to compare with the following implementation? https://github.com/sappelhoff/pyprep/blob/master/pyprep/removeTrend.py#L92 They use `window = (1.5/freq)*sample_rate` not `window = sample_rate / freq`


>     for consistency as the other 2 methods (polynomial and standardized) are using 0.5Hz for lowcut, so it should locreg
Agreed, just want to make sure we compute that correctly, and I think since `sampling_rate` is now an argument for `signal_detrend()` we should include `sampling_rate` as an input rather than computing the `window` and `stepsize` with `sampling_rate`.

danibene:
@vasileermicioi update: I tried out various definitions of the window with a simulated signal: https://gist.github.com/danibene/e3ed1200a75ea72a3c52bfb0ac248a2c

Now I'm also wondering if the [pyprep implementation that I linked earlier ](https://github.com/sappelhoff/pyprep/blob/master/pyprep/removeTrend.py#L92)is incorrect. Any insights you have here would be appreciated!

In any case, I agree that the documentation should be updated so that it's clear that the lowcut is not equivalent to the window.

vasileermicioi:
> Now I'm also wondering if the [pyprep implementation that I linked earlier ](https://github.com/sappelhoff/pyprep/blob/master/pyprep/removeTrend.py#L92)is incorrect. Any insights you have here would be appreciated!

yes, it is wrong too, because taking a window with 50% larger will have the effect of reducing the frequency by 1/3, for example if we want to remove all frequencies below 0.5Hz, it will remove all below 0.33Hz instead, and if we want to remove below 2Hz, it will remove only below 1.34Hz

neurokit2/signal/signal_detrend.py
```diff
@@ -43,6 +43,9 @@ def signal_detrend(
         Only used if ``method`` is "loess". The parameter which controls the degree of smoothing.
     window : float
         Only used if ``method`` is "locreg". The detrending ``window`` should correspond to the
+        sample rate divided by the desired low frequency band to remove (window=sample_rate / frequency) .
+        For instance, to remove frequencies below 0.67Hz for a signal sampled at 1000Hz, a window
+        of the size round(1000 / 0.67) = ``1.5*1000`` should be used.
         desired low frequency band to remove (for instance, ``1.5`` will remove frequencies below
```
DominiqueMakowski, line 49:
```suggestion
```

neurokit2/signal/signal_detrend.py
```diff
@@ -43,6 +43,9 @@ def signal_detrend(
...
         1.5Hz).
```
DominiqueMakowski, line 50:
```suggestion
```

neurokit2/signal/signal_detrend.py
```diff
@@ -43,8 +43,9 @@ def signal_detrend(
         Only used if ``method`` is "loess". The parameter which controls the degree of smoothing.
     window : float
         Only used if ``method`` is "locreg". The detrending ``window`` should correspond to the
-        desired low frequency band to remove (for instance, ``1.5`` will remove frequencies below
-        1.5Hz).
+        sample rate divided by the desired low frequency band to remove (window=sample_rate / frequency) .
+        For instance, to remove frequencies below 0.67Hz for a signal sampled at 1000Hz, a window
+        of the size round(1000 / 0.67) = ``1.5*1000`` should be used.
```
danibene, line 48: With the current implementation in which `window` is multiplied by `sampling_rate`, I think the correct input for this example would be `(1000/0.67)/1000` rather than `round(1000 / 0.67)`. 

@DominiqueMakowski we could 
1. change the doc example to reflect the current implementation, 
2. change the implementation so that `window` isn't multiplied by `sampling_rate`, or 
3. change the implementation so that `lowcut` is an argument instead of `window` and compute `window` from `sampling_rate / lowcut` (assuming that the [pyprep implementation](https://github.com/sappelhoff/pyprep/blob/master/pyprep/removeTrend.py#L92) is actually incorrect, which I don't feel qualified to judge honestly)

danibene (changes requested):

neurokit2/signal/signal_detrend.py
```diff
@@ -43,6 +43,9 @@ def signal_detrend(
         Only used if ``method`` is "loess". The parameter which controls the degree of smoothing.
     window : float
         Only used if ``method`` is "locreg". The detrending ``window`` should correspond to the
+        sample rate divided by the desired low frequency band to remove (window=sample_rate / frequency) .
+        For instance, to remove frequencies below 0.67Hz for a signal sampled at 1000Hz, a window
+        of the size round(1000 / 0.67) = ``1.5*1000`` should be used.
```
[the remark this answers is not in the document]
danibene: Sorry, this was an old comment, didn't realize it was "pending", see new comment below

neurokit2/signal/signal_detrend.py
danibene, line 48:
```suggestion
        of the size (1000 / 0.67)/1000 should be used.
```

studies/ecg_benchmark/make_data.py
```diff
@@ -152,7 +152,7 @@ def standardization(ecg, sampling_rate):
 #def locreg(ecg, sampling_rate):
 #    ecg = nk.signal_detrend(ecg,
 #                            method="locreg",
-#                            window=0.5*sampling_rate,
+#                            window=2*sampling_rate,
```
danibene, line 155:
```suggestion
#                            window=(sampling_rate/0.5)/sampling_rate,
```

studies/ecg_benchmark/make_data.py
```diff
@@ -152,7 +152,7 @@ def standardization(ecg, sampling_rate):
...
 #                            stepsize=0.02*sampling_rate)
```
danibene, line 156:
```suggestion
#                            stepsize=0.02)
```

vasileermicioi:
@danibene now the `window` parameter is not a window anymore, but is the detrend frequency, as `(sample_rate / freq) / sample_rate` is the same as `freq`, so perhaps the parameter name (window) should be changed to detrend_frequency

danibene:
> @danibene now the `window` parameter is not a window anymore, but is the detrend frequency, as `(sample_rate / freq) / sample_rate` is the same as `freq`, so perhaps the parameter name (window) should be changed to detrend_frequency

Wouldn't it be `1/freq`? But we could change the implementation so that it's `freq`

vasileermicioi:
> > @danibene now the `window` parameter is not a window anymore, but is the detrend frequency, as `(sample_rate / freq) / sample_rate` is the same as `freq`, so perhaps the parameter name (window) should be changed to detrend_frequency
> 
> Wouldn't it be `1/freq`? But we could change the implementation so that it's `freq`

yes, you are right

danibene:
@DominiqueMakowski any thoughts on whether we should deviate from the pyprep implementation?

DominiqueMakowski:
pyprep is supposedly pretty robust, perhaps it's worth opening an issue on their repo to get their thoughts?

vasileermicioi:
@danibene 
I am not sure I understand completely the answer, but to me, it sounds like "you are right, but 0.67Hz is a good frequency for detrending"
https://github.com/VisLab/EEG-Clean-Tools/issues/33

DominiqueMakowski:
> you are right, but 0.67Hz is a [reasonable] frequency for EEG detrending

That's my understanding as well..., but we can wait 'til Dani comes back :)

danibene:
@vasileermicioi based on the responses we received: https://github.com/VisLab/EEG-Clean-Tools/issues/33#issuecomment-1304067303 it looks like you're correct, but they don't plan to change the implementation.

How about we keep the `window` parameter as is so that our implementation is still based off of prep's, but just update our documentation that `window` should be 1/cutoff? If you agree could you please make those changes? Then I think we can merge.

DominiqueMakowski:
Yeah that seem like a good resolve, this way we don't change the default as compared to prep but provide more context and info for users to potentially fine tune it

DominiqueMakowski:
Maybe we can make the docs change and close the PR (I'd like to merge the dev branch soon :)

DominiqueMakowski:
> Maybe we can make the docs change and close the PR

I was thinking "merge" the PR actually 😢 @vasileermicioi could you reopen one?

Issue #744, opened 2022-11-10
ECG R-Peak detection using visibility graphs


sokolmarek:
**Describe the solution you'd like**
I recently came across a new algorithm for detecting R peaks that I found very interesting and I haven't seen it anywhere before. The algorithm uses a time series to visibility graph transformation. The cool thing is that it has linear complexity and it could also be useful for preprocessing in HRV analyses, as it is more robust to outliers. The authors benchmarked the method on the GUDB database against other SWTs (e.g. Kalidas 2017) and matched filter QRS detectors. The performance difference compared between the methods seems quite diametric to me. So I thought it was interesting enough that it would be nice to implement this method within the Neurokit2 library. Below I attach a reference to the corresponding article.

- [T. Koka and M. Muma, "Fast and Sample Accurate R-Peak Detection for Noisy ECG Using Visibility Graphs," 2022 44th Annual International Conference of the IEEE Engineering in Medicine & Biology Society (EMBC), 2022, pp. 121-126, doi: 10.1109/EMBC48229.2022.9871266.](https://ieeexplore.ieee.org/document/9871266)

**How could we do it?**
I have already implemented the method for my own purposes with the compatibility of the latest version of the [ts2vg](https://pypi.org/project/ts2vg/) library. In the attached article they have everything nicely described. Eventually, I would create a pull request and redo my implementation according to neurokit conventions:)

DominiqueMakowski:
Hi Marek! 

That's very interesting, I wasn't aware of the visibility graph concept. Exciting!
By all means, do open a PR, and we'll assist you with the implementation :)

PR #745, opened 2022-11-11
[Feature] ECG R-Peak detection using visibility graphs 

sokolmarek:
# Description

This PR aims at adding the feature mentioned here #744. 

# Proposed Changes

I implemented a method for R peak detection that uses an ECG to visibility graph transformation according to this [article](https://ieeexplore.ieee.org/document/9871266).

# Checklist

### DIFF
--- a/neurokit2/ecg/ecg_findpeaks.py
+++ b/neurokit2/ecg/ecg_findpeaks.py
@@ -1075,13 +1075,7 @@ def _ecg_findpeaks_vgraph(signal, sampling_rate=1000, lowcut=3, order=2, **kwarg
             " this method to run. Please install it first (`pip install ts2vg`)."
         ) from import_error
 
-    # Filter signal using highpass Butterworth filter
-    nyq = 0.5 * sampling_rate
-    high = lowcut / nyq
-    b, a = scipy.signal.butter(order, high, btype="highpass")
-    filtered = scipy.signal.filtfilt(b, a, signal)
-
-    N = len(filtered)
+    N = len(signal)
     M = 2 * sampling_rate
     w = np.zeros(N)
     rpeaks = []
@@ -1095,7 +1089,7 @@ def _ecg_findpeaks_vgraph(signal, sampling_rate=1000, lowcut=3, order=2, **kwarg
     n_segments = int((N - deltaM) / (M - deltaM)) + 1
 
     for segment in range(n_segments):
-        y = filtered[L:R]
+        y = signal[L:R]
 
         # Compute the adjacency matrix to the directed visibility graph
         A = ts2vg.NaturalVG(directed="top_to_bottom").build(y).adjacency_matrix()
@@ -1122,7 +1116,7 @@ def _ecg_findpeaks_vgraph(signal, sampling_rate=1000, lowcut=3, order=2, **kwarg
             R = N
 
         # Multiply signal by its weights and apply thresholding algorithm
-        weighted_signal = filtered * w
+        weighted_signal = signal * w
         rpeaks = _ecg_findpeaks_peakdetect(weighted_signal, sampling_rate)
         rpeaks = np.array(rpeaks, dtype="int")
     return rpeaks
--- a/neurokit2/ecg/ecg_findpeaks.py
+++ b/neurokit2/ecg/ecg_findpeaks.py
@@ -94,6 +94,8 @@ def _ecg_findpeaks_findmethod(method):
         return _ecg_findpeaks_WT
     elif method in ["rodrigues2020", "rodrigues2021", "rodrigues", "asi"]:
         return _ecg_findpeaks_rodrigues
+    elif method in ["vg", "vgraph", "koka2022"]:
+        return _ecg_findpeaks_vgraph
     elif method in ["promac", "all"]:
         return _ecg_findpeaks_promac
     else:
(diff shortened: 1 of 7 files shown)


neurokit2/ecg/ecg_findpeaks.py
```diff
@@ -1051,6 +1053,80 @@ def _ecg_findpeaks_rodrigues(signal, sampling_rate=1000, **kwargs):
...
+        raise ImportError(
+            "NeuroKit error: ecg_findpeaks(): the 'ts2vg' module is required for"
+            " this method to run. Please install it first (`pip install ts2vg`)."
+        ) from import_error
+
+    # Filter signal using highpass Butterworth filter
+    nyq = 0.5 * sampling_rate
+    high = lowcut / nyq
+    b, a = scipy.signal.butter(order, high, btype="highpass")
+    filtered = scipy.signal.filtfilt(b, a, signal)
```
DominiqueMakowski, line 1082: Should we filter the signal again or do we assume that this cleaning should be done beforehand (using `ecg_clean()`) and the signal provided should be clean?

neurokit2/ecg/ecg_findpeaks.py
```diff
@@ -1051,6 +1053,80 @@ def _ecg_findpeaks_rodrigues(signal, sampling_rate=1000, **kwargs):
...
+    References
+    ----------
+    - T. Koka and M. Muma (2022), Fast and Sample Accurate R-Peak Detection for Noisy ECG Using
+      Visibility Graphs. In: 2022 44th Annual International Conference of the IEEE Engineering
+      in Medicine & Biology Society (EMBC). Uses the Pan and Tompkins thresholding.
+
+    """
+    # Try loading ts2vg
+    try:
+        from ts2vg import NaturalVG
```
DominiqueMakowski, line 1071: just for clarity, I prefer to have the name of the package as a prefix to its function calls, it helps me to remind from where each function comes

```suggestion
        import ts2vg
```

neurokit2/ecg/ecg_findpeaks.py
```diff
@@ -1051,6 +1053,80 @@ def _ecg_findpeaks_rodrigues(signal, sampling_rate=1000, **kwargs):
...
+
+    # Compute number of segments
+    deltaM = int(gamma * M)
+    n_segments = int((N - deltaM) / (M - deltaM)) + 1
+
+    for segment in range(n_segments):
+        y = filtered[L:R]
+
+        # Compute the adjacency matrix to the directed visibility graph
+        A = NaturalVG(directed="top_to_bottom").build(y).adjacency_matrix()
```
DominiqueMakowski, line 1101:
```suggestion
        A = ts2vg.NaturalVG(directed="top_to_bottom").build(y).adjacency_matrix()
```

DominiqueMakowski:
Looks great, thanks! Please consider adding yourself to the [contributors](https://github.com/neuropsychology/NeuroKit/blob/master/AUTHORS.rst) :)

neurokit2/ecg/ecg_findpeaks.py
sokolmarek: Well, I did it like for example with the Kalidas, Christov, Zong, etc. detectors where nothing is filtered in ecg_clean() and the raw signal is passed on. This means that there is no re-filtering and filtering is only part of the detector.

danibene: @sokolmarek Unless this is based on an existing implementation (I saw the paper mentioned that the Python code would be made publicly available on GitHub but I couldn't find it), I think it would be better to add a new method to `ecg_clean()` using the `"butterworth"` method of `signal_filter()`

danibene (commented):
This looks good to me! I would consider adding a clean method before merging (see my other comment)

neurokit2/ecg/ecg_findpeaks.py
danibene: And @DominiqueMakowski I also think we should consider doing the same for some of the other detectors, at least Zong where it is also a butterworth filter: https://github.com/neuropsychology/NeuroKit/blob/5c749d514ad9c9d5fd6d65c39c9a101e1eb3e4cf/neurokit2/ecg/ecg_findpeaks.py#L476-L482

However, the `"butterworth"` method of `signal_filter()` uses zero-phase filtering with [`scipy.signal.sosfiltfilt()`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.sosfiltfilt.html) whereas the current Zong implementation uses [`scipy.signal.lfilter()`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.lfilter.html), not sure if another method should be added to `signal_filter()` so that no functional changes are made?

DominiqueMakowski: > we should consider doing the same for some of the other detectors, at least Zong where the cleaning is also just a butterworth filter

Yes that makes sense, as otherwise we risk doing double cleaning and cannot dissociate the effect of cleaning method vs peak detection per se

sokolmarek: @danibene So I added a new `ecg_clean()` method using `signal_filter()` and removed the original filtering from the detector. At the same time, I think it's a good idea to remove the other filtering in other detectors in the future. Signal preprocessing is often individual and one has to play with it sometimes to design the right filter. There is no gold standard. The fact that such filtering worked for their ECG recordings within the Zong detector does not mean that it will be as effective elsewhere.

danibene: > So I added a new ecg_clean() method using signal_filter() and removed the original filtering from the detector.

@sokolmarek Thank you! `signal_filter()` takes `lowcut` in Hz so you can remove the normalization step and directly use `lowcut=4`. Also, is there a reason you want to use `"butterworth_ba"` instead of `"butterworth"` (which uses [`scipy.signal.sosfiltfilt()`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.sosfiltfilt.html))? I think we have generally had better results with the latter

sokolmarek: > > So I added a new ecg_clean() method using signal_filter() and removed the original filtering from the detector.
> 
> @sokolmarek Thank you! `signal_filter()` takes `lowcut` in Hz so you can remove the normalization step and directly use `lowcut=4`. Also, is there a reason you want to use `"butterworth_ba"` instead of `"butterworth"` (which uses [`scipy.signal.sosfiltfilt()`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.sosfiltfilt.html))? I think we have generally had better results with the latter

@danibene All right, I'll edit it:) Since there was no mention of using the second-order section in the article as part of the filtering, I didn't use it here either. I'll rewrite it to sos, but I think in this low order filter it won't have that much effect. At least it minimizes the ringing effect on the edges I guess...

danibene: @sokolmarek thank you :) I'll go ahead and merge this now, I think any changes to the other peak detection functions e.g., adding a new `ecg_clean()` method for the Zong detector, should be done in a separate PR

PR #746, opened 2022-11-13
[Feature] Compute RSP cycle symmetry features + new amplitude method

DominiqueMakowski:


DominiqueMakowski:
@rostro36 FYI I included RVT computation by default in `rsp_process()` + changed the default RVT method to Harrison (as it was faster), if you think that's not a good idea let me know!

rostro36:
@DominiqueMakowski Thanks for brushing up older code!

Unfortunately, I am no expert in the field, but as much as I read, the Birn method is the most basic and straight forward method, while still being the main benchmark for both other papers. Power's method is an adaptation of it, whereas Harrison does quite a few things different.

Personally, I would go for the Power method as default, as it is a small adaption (hopefully improvement) of the standard Birn method. If that does not work, or the user is aware of Harrison, they can opt for that and also think about the additional parameters. But I have not timed the functions yet and don't know how big the improvement would be on a standard task. Maybe adding a small comment next to Harrison (fastest) would be a middle ground.

Let me know if you need some support; I should be able to make some time on the weekend if needed.

PR #750, opened 2022-11-16
[Fix] RVT improvements

DominiqueMakowski:


DominiqueMakowski:
Follow up to https://github.com/neuropsychology/NeuroKit/pull/746#issuecomment-1315867089

@rostro36 said

> Unfortunately, I am no expert in the field, but as much as I read, the Birn method is the most basic and straight forward method, while still being the main benchmark for both other papers. Power's method is an adaptation of it, whereas Harrison does quite a few things different.
> 
> Personally, I would go for the Power method as default, as it is a small adaption (hopefully improvement) of the standard Birn method. If that does not work, or the user is aware of Harrison, they can opt for that and also think about the additional parameters. But I have not timed the functions yet and don't know how big the improvement would be on a standard task. Maybe adding a small comment next to Harrison (fastest) would be a middle ground.
> 
> Let me know if you need some support; I should be able to make some time on the weekend if needed.

my only issue with "power" as the default is that it's significantly slower and I wouldn't want users to be surprised by long computation times when then run `rsp_process()`, which is meant as a default hassle-free "one ring to bind them all" kind of function. 

Some things that I can think of include:

- [ ] Have a different default in `rsp_rvt()` (power) and `rsp_process()` (birn or harrison)
- [ ] Maybe we could profile power method and see if there is some possibilities of making it more efficient?
- [ ] For Birn: can we change the interpolation step by using `nk.signal_interpolate()` and expose the interpolation method? This would allow for a smoother RVT which I suppose is more plausible?
- [ ] Critically, we should aim at allowing users to change the RVT method in `rsp_process()` (as well as other methods): the best way to do it is to implement **reports** to RSP, same as we did for PPG. This is a bigger endeveaour that we could split into two steps 
  - [ ] Add a [rsp_methods()](https://github.com/neuropsychology/NeuroKit/blob/master/neurokit2/ppg/ppg_methods.py) that sanitizes and dispatches all the different methods (for cleaning, for peak detection, for RVT etc), with an empty report text for now
  - [ ] Later, add the report content

neurokit2/rsp/rsp_rvt.py
```diff
@@ -139,6 +139,8 @@ def _rsp_rvt_birn(
         ending_peak = peak_coords[ending_peak_index]
         mid_peak.append(round((starting_peak + ending_peak) / 2))
         seconds_delta.append((ending_peak - starting_peak) / sampling_rate)
+
+    # Interpolate
     output_range = range(len(zsmooth_signal))
     rvt_time = np.interp(output_range, mid_peak, seconds_delta)
```
DominiqueMakowski, line 145: Can we change these steps to using nk.signal_interpolate()?

danibene:
> - [ ] Critically, we should aim at allowing users to change the RVT method in `rsp_process()` (as well as other methods): the best way to do it is to implement **reports** to RSP, same as we did for PPG. This is a bigger endeveaour that we could split into two steps 
>   - [ ] Add a [rsp_methods()](https://github.com/neuropsychology/NeuroKit/blob/master/neurokit2/ppg/ppg_methods.py) that sanitizes and dispatches all the different methods (for cleaning, for peak detection, for RVT etc), with an empty report text for now
>   - [ ] Later, add the report content
> 

@DominiqueMakowski 

This makes sense to me! Let me know if you need help adapting `ppg_methods()`, I have limited time the next couple of weeks but happy to support when I can

PR #766, opened 2023-01-25
Fix ms hrv rsa + proposed fix for RSA P2T

vivile42:
Convert RSA P2T output in ms
# Description
In the description it says it should return ms but it was in second i've fixed it as discussed in [here](https://github.com/neuropsychology/NeuroKit/discussions/763)

I've also reworked the p2t to compute the variation between the max RRI in exh and the min RRI in inh following the [paper](https://pubmed.ncbi.nlm.nih.gov/2100356/) from Grossman 1990  

This is a working version (on the data i've tested at least )  but should just be a starting point since it's not really validated i hope it can be helpful :) 

# Proposed Changes
- modified the _hrv_rsa_p2t function multiplying by 1000 the difference in TF after it has been divided by the sampling rate
- added cycle inh and exh separately to compute minima
- added buffer of 750 ms to the peak and end cycle in order to take into consideration end of following RRI (see Grossman 1990) to me it makes sense but i don't know if it's still done like that ( i couldn't find recent code to double check)  
- when max RRI exh - min RRI inh is negative the breathing cycle is counted as zero (again same procedure from Grossman 1990)
- important: in that paper they take into account only breathing cycles in which the minima fell into inhalation and was preceded by a longer RRI interval, similarly the max has to be preceded by a a shorter interval. I don't want this for my data since it's an active condition in which people breathe fast, i don't want to have too much missing data. You should discuss whether you want to implement this part or not, maybe it can be toggled as optional ? 

# Checklist

### DIFF
--- a/neurokit2/hrv/hrv_rsa.py
+++ b/neurokit2/hrv/hrv_rsa.py
@@ -215,19 +215,29 @@ def _hrv_rsa_p2t(
     """Peak-to-trough algorithm (P2T)"""
 
     # Find all RSP cycles and the Rpeaks within
-    cycles_rri = []
+    cycles_rri_inh = []
+    cycles_rri_exh = []
     for idx in range(len(rsp_onsets) - 1):
         cycle_init = rsp_onsets[idx]
+        rsp_peak=rsp_peaks[idx]
         cycle_end = rsp_onsets[idx + 1]
-        cycles_rri.append(rpeaks[np.logical_and(rpeaks >= cycle_init, rpeaks < cycle_end)])
+    # separately select RRI for inhalation and exhalation    
+        cycles_rri_inh.append(rpeaks[np.logical_and(rpeaks >= cycle_init, rpeaks < rsp_peak)])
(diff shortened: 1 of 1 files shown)


DominiqueMakowski:
Hi @vivile42, sorry for the delay I was relocating to another country. I will merge as the check errors seems unrelated to these changes, thanks a lot!

DominiqueMakowski:
>  in that paper they take into account only breathing cycles in which the minima fell into inhalation and was preceded by a longer RRI interval, similarly the max has to be preceded by a a shorter interval. I don't want this for my data since it's an active condition in which people breathe fast, i don't want to have too much missing data. You should discuss whether you want to implement this part or not, maybe it can be toggled as optional ?

Interesting, yes we could add am argument to toggle that on or off. Feel free to open a new PR (so sorry I should have waited instead of merging this one!). 

(and consider adding yourself to the [contributors](https://github.com/neuropsychology/NeuroKit/blob/dev/AUTHORS.rst))

Issue #701, opened 2022-09-06
Additional PPG peak detectors


peterhcharlton:
**The suggestion**
Currently there is one PPG peak detection algorithm available in the [ppg_findpeaks](https://neuropsychology.github.io/NeuroKit/functions/ppg.html#ppg-findpeaks) function, whereas there are several peak detection algorithms available in the [ecg_peaks](https://neuropsychology.github.io/NeuroKit/functions/ecg.html#ecg-peaks) function. A recent [publication](https://iopscience.iop.org/article/10.1088/1361-6579/ac826d) _(disclaimer: written by my colleagues and I)_ found that several PPG peak detection algorithms have been proposed in the literature, and identified the best-performing algorithms (see [here](https://paperswithcode.com/sota/photoplethysmography-ppg-beat-detection-on) for a quick summary). Therefore, I'd suggest implementing additional PPG peak detection algorithms - perhaps including those which have been identified as the best performing.

**How could we do it?**
The recently released [ppg-beats Matlab toolbox](https://ppg-beats.readthedocs.io/) _(disclaimer: written by my colleagues and I)_ contains Matlab implementations of several PPG peak detection algorithms. We could translate the relevant algorithms from this Matlab code into Python code for inclusion in the [ppg_findpeaks](https://neuropsychology.github.io/NeuroKit/functions/ppg.html#ppg-findpeaks) function. I'd be happy to work on this if helpful.

DominiqueMakowski:
I think that'd be super helpful, actually, as NK's PPG-related functions are a bit under-developped (as compared to ECG). + It will be a good way to let NK users discover your work. Feel free to open a PR and don't hesitate to credit yourself & your team; and your own matlab software where it's due ;) (for example, in the references, the description of the function, as comments in the code etc.)

peterhcharlton:
Great, thanks, will do. This will take me a bit of time. I'll aim to contribute a PPG beat detector in the next few weeks.

LegrandNico:
I was planning to translate the Multi-Scale Peak and Trough Detection algorithm for [Systole](https://github.com/embodied-computation-group/systole) in the next weeks from your Matlab code @peterhcharlton , so I am happy to help with this one for NeuroKit 👍

peterhcharlton:
[...]

PR #768, opened 2023-01-26
Added MSPTD ppg peak detector

peterhcharlton:
# Description

This PR aims to add an additional PPG peak detector, as discussed in Issue #701 . The new peak detector is the 'MSPTD' method, and implementation of:

_Bishop SM, Ercole A (2018) Multi-scale peak and trough detection optimised for periodic and quasi-periodic neuroscience data. doi:10.1007/978-3-319-65798-1_39._

# Proposed Changes

- I changed the `ppg_findpeaks()` function to include the 'msptd' method. Please note that the method is designed for short signals of relatively low sampling frequency. I'd be happy to adapt it to overcome these shortcomings either now or in the future.
- I changed `tests_ppg.py` to include an additional test for the 'msptd' method.

# Checklist

_Please note: I haven't run the checks - I'll be happy to review any errors that arise when the automated checks are run._


DominiqueMakowski:
Awesome, thanks a lot!! I have edited the examples but just from the plot I wonder if we should not `-1` all the onsets indices

image

Also, I added a note to create a `ppg_peaks()` master function that would also return onsets and other info similar to ecg_peaks etc.

peterhcharlton:
I've made a slight adjustment, which I think resolves the issue of the onsets being offset from the true onsets in the plot above:
image

I agree with creating a `ppg_peaks()` master function. We could use this function to provide onsets (or peaks) when using beat detectors that only provide one or the other (and not both). E.g. the Elgendi peak detector currently only provides peaks, and we could implement a step to find onsets (as the minima between consecutive peaks) if they are not provided by the peak detector.

DominiqueMakowski:
Great, I'll merge that and we can then start working on the peaks master function to provide more info

Issue #772, opened 2023-02-02
Strange indexing error using "ecg_process"


niwolpert:
**Bug description**
I'm calling 'ecg_process' on an epochs of ECG signals of 5000 samples each (sample frequency = 250 Hz). With most epochs this works fine, but with some epochs, it results in the error "KeyError: '[5000] not in index'" (here 5000 corresponds to the number of samples, if I change the epoch length, it throws the same error). I cannot explain why it throws this error with some epochs. There are no Nan in the epochs and the ECG signal looks clean and regular.

![neurokit error](https://user-images.githubusercontent.com/50142717/216299756-fcd2574b-3f6c-4000-bf78-fb33d5c9cf5a.png)

**To Reproduce**

```
import pandas as pd
import neurokit2 as nk
ecg_signal = pd.read_csv('ecg_signal.csv')        # load csv provided in attachment
df_raw, info = nk.ecg_process(ecg_signal, sampling_rate=250)
```

**Expected behaviour**
I expect the function to return me the raw df and info dictionary.

**System Specifications**

- OS: Windows (WindowsPE 64bit) 
- Python: 3.9.12 
- NeuroKit2: 0.2.0 
- NumPy: 1.22.3 
- Pandas: 1.4.4 
- SciPy: 1.7.3 
- sklearn: 1.0.2 
- matplotlib: 3.5.2

[ecg_signal.csv](https://github.com/neuropsychology/NeuroKit/files/10567836/ecg_signal.csv)

danibene:
Hi, thanks for pointing this out!

I think this is happening because `ecg_delineate()` detects points in the segmented ECG, and since the last heartbeat segment has zeros appended at the end (I think that's done so that the length of that last segment isn't shorter than the previous segments), it is possible that the point indices detected don't actually correspond to indices in the original ECG signal. One possible way to deal with this would be to remove the appended zeros from the final heartbeat:
https://github.com/danibene/NeuroKit/commit/1bedb50dc7dc65f51533fcc0b6cefecc5da1a722

Or another way could be to go through all the points detected in `ecg_delineate()` and remove any indices that are higher than the last index in the original signal.

Any thoughts @DominiqueMakowski ?

DominiqueMakowski:
> Or another way could be to go through all the points detected in ecg_delineate() and remove any indices that are higher than the last index in the original signal.

Yes, I think we could add that by default in ecg_delinate() as a sanity postprocess where it removes all invalid indices

[...]

PR #777, opened 2023-02-13
[Fix] invalid indices in ecg_delineate

danibene:
# Description

This prevents indices larger than the ECG signal indices from being returned by `ecg_delineate()`. See https://github.com/neuropsychology/NeuroKit/issues/772

# Proposed Changes

I changed `ecg_delineate()` so that regardless of which delineation method is used, the output is postprocessed such that any indices that are larger than the ECG signal indices are removed (originally this postprocessing was only done in `_ecg_delineator_peak()`.

I also changed `ecg_segment()` so that the final heartbeat segment only contains points from the original ECG signal (not zeros appended at the end), since that caused invalid indices to be detected within `ecg_delineate()`. But I can revert this change or make it an optional argument, as invalid indices will be removed in `ecg_delineate()` anyway.

# Checklist

### DIFF
--- a/neurokit2/ecg/ecg_delineate.py
+++ b/neurokit2/ecg/ecg_delineate.py
@@ -164,6 +164,11 @@ def ecg_delineate(
             "NeuroKit error: ecg_delineate(): 'method' should be one of 'peak'," "'cwt' or 'dwt'."
         )
 
+    # Ensure that all indices are not larger than ECG signal indices
+    for _, value in waves.items():
+        if value[-1] >= len(ecg_cleaned):
+            value[-1] = np.nan
+
     # Remove NaN in Peaks, Onsets, and Offsets
     waves_noNA = waves.copy()
     for feature in waves_noNA.keys():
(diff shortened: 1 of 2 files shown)


DominiqueMakowski:
Thanks a lot Dani! seems like there is an error in the execution fo this notebook: docs\examples\bio_custom\bio_custom.ipynb I'm not sure it's related to that particular PR but just in case could you try tu run it manually to make sure it's unrelated?

DominiqueMakowski:
Maybe it creates a problem in the plot of the segments?

danibene:
> Thanks a lot Dani! seems like there is an error in the execution fo this notebook: docs\examples\bio_custom\bio_custom.ipynb I'm not sure it's related to that particular PR but just in case could you try tu run it manually to make sure it's unrelated?

How do you usually run the notebook to test a specific version of NeuroKit?

I added this cell at the top of the notebook (using the `fix/ecg_segment_extra_samples` branch) and it ran fine:

```
import pathlib
import sys

package_dir = pathlib.Path().resolve().parent.parent.parent # where NeuroKit folder is located
sys.path.append(str(package_dir))
```

DominiqueMakowski:
> to test a specific version of NeuroKit?

Actually that's a good question to which I didn't had answer.  Till now I pip uninstalled and installed the target version... 

the workflow error is unrelated to this PR is it?

DominiqueMakowski:
I just re-run and all is in order.

> so that the final heartbeat segment only contains points from the original ECG signal (not zeros appended at the end

I am not sure about this, I like the idea of returning same-length epochs. Maybe we should pad with nan instead?

Issue #780, opened 2023-02-14
eda/_eda_findpeaks_vanhalem2020 function does not calculate Skin conductance height correctly


GanshengT:
The recovery of skin conductance response is miscalculated in eda_peaks() when using vanhalem2020 method. I think the problem is that the height of the skin conductance response is miscalculated or defined incorrectly in eda/_eda_findpeaks_vanhalem2020.

**Describe the bug**
    info = {
        "SCR_Onsets": info["Onsets"][idx],
        "SCR_Peaks": info["Peaks"][idx],
        "SCR_Height": info["Height"][idx],
    }
info["Height"][idx] is calculated with nk.signal_findpeak(). I think in eda_peaks() function, the offset/recovery of skin conductance response is based on the peak value of each peak, instead of the height of the peak defined in nk.signal_findpeak(). 

**To Reproduce**
eda_signal = nk.data("bio_eventrelated_100hz")["EDA"]
_, vanhalem2020 = nk.eda_peaks(data2plot['cda_phasic'], method="vanhalem2020", sampling_rate=100)

**Expected behaviour**
I think "SCR_Height": eda_phasic[peaks["Peaks"]] might fix the problem.

**Screenshots**
purple denotes the onsets from vanhalem2020, and red lines denote the recovery/offset.
<img width="583" alt="Screen Shot 2023-02-14 at 4 48 33 PM" src="https://user-images.githubusercontent.com/49130176/218880731-9b1f3dfe-756a-42e3-a93f-b44ea7030e00.png">

GanshengT:
Mentioned issue has been fixed. There is another concern left for discussion: in the current script, if eda signals ends before the peak recovers, the offset will not be detected.

Issue #779, opened 2023-02-14
function _eda_findpeaks_nabian2018 does not follow the reference paper


GanshengT:
Hi all,
 _eda_findpeaks_nabian2018 aims at identify skin conductance response. The algorithm can be found in the following paper:
https://ieeexplore.ieee.org/document/8510820

**Describe the bug**
SCRs contained within the EDA signal can be detected by performing differentiation and subsequent convolution with a 20-point Bartlett window. The differentiation step is missed in the function in neuroKit2.

**To Reproduce**
_, nabian2018 = nk.eda_peaks(eda_phasic, method="nabian2018")
The original code is:
'''
    # smooth
    eda_phasic = signal_smooth(eda_phasic, kernel="bartlett", size=20)

    # zero crossings
    pos_crossings = signal_zerocrossings(eda_phasic, direction="positive")
    neg_crossings = signal_zerocrossings(eda_phasic, direction="negative")
'''

**Expected behaviour**
To find the zero crossing, we need to perform differentiation first. The fix would be simple. We can add eda_phasic = np.diff(eda_phasic) before smoothing the signal.

Issue #778, opened 2023-02-14
eda_sympathetic() assumes the sampling frequency of 400Hz


GanshengT:
Hi NK community,

I found the eda_sympathetic() function assumes the sampling frequency is 400Hz when calculating sympathetic index from EDA signal. In the original paper, the authors say "Before down-sampling, the data was filtered with an 8th-order Chebyshev Type I low-pass filter (0.8 Hz). Down-sampling from 400 to 2 Hz was performed in two steps (using consecutive down-sampling factors of 1/20 and 1/10, respectively). Finally, signals were high-pass filtered (0.01 Hz, Butterworth, 8th order) to remove any trend."

eda_sympathetic() calls _eda_sympathetic_posada to calculate the sympathetic index:
elif method.lower() in ["posada", "posada-quintero", "quintero"]:
        out = _eda_sympathetic_posada(eda_signal, frequency_band=frequency_band, show=show)
    
When looking at _eda_sympathetic_posada, you will find signal was downsampled twice (1/200). The problem is that the signal_filter assumes the sampling rate of the downsampled signal is 2.

def _eda_sympathetic_posada(eda_signal, frequency_band=[0.045, 0.25], show=True, out={}):

    # First step of downsampling
    downsampled_1 = scipy.signal.decimate(eda_signal, q=10, n=8)  # Keep every 10th sample
    downsampled_2 = scipy.signal.decimate(downsampled_1, q=20, n=8)  # Keep every 20th sample

    # High pass filter
    eda_filtered = signal_filter(
        downsampled_2, sampling_rate=2, lowcut=0.01, highcut=None, method="butterworth", order=8
    )

Reference:
Posada-Quintero, H. F., Florian, J. P., Orjuela-Cañón, A. D., Aljama-Corrales, T.,
      Charleston-Villalobos, S., & Chon, K. H. (2016). Power spectral density analysis of
      electrodermal activity for sympathetic function assessment. Annals of biomedical engineering,
      44(10), 3124-3135.

**To Reproduce**
eda = nk.data('bio_resting_8min_100hz')['EDA']
indexes_posada = nk.eda_sympathetic(eda, sampling_rate=100, method='posada', show=True)

**Expected behaviour**
The eda signal is sampled at 100Hz. The high pass filter would need the input sampling_rate = sampling_rate/200, which is 0.5

**Screenshots**
If applicable and useful, you can add screenshots by drag-and-dropping the images in the issue window.

**System Specifications**

[...]

PR #783, opened 2023-02-16
[Fix] EDA peak detection and sympathetic activity

GanshengT:
# Description

This PR aims at fix the issue mentioned at #778, #780 and #779

# Proposed Changes

if the user chooses to 'posada' to calculate Sympathetic Nervous System Index, EDA signal will be resampled at 400Hz. A low-pass filter is also added according to original paper. `eda_findpeaks()` using methods proposed in nabian2018 is reviewed and improved. Differentiation has been added before smoothing. Skin conductance response criteria have been revised based on the original paper.

In the eda.find_peaks sub-module, _eda_findpeaks_vanhalem2020 has been revised. The algorithm is changed based on the orginal paper.
` We specified a peak when a consistent increase of 0.5 seconds would follow a consistent decrease of 0.5 seconds.`
However, if eda signals ends before the peak recovers, the offset will not be detected.

# Checklist

- [+ ] I have read the [CONTRIBUTING](https://github.com/neuropsychology/NeuroKit/blob/master/.github/CONTRIBUTING.rst#structure-and-code) file.
- [ +] My PR is targetted at the **dev branch** (and not towards the master branch).
- [ +] I ran the [CODE CHECKS](https://github.com/neuropsychology/NeuroKit/blob/master/.github/CONTRIBUTING.rst#run-code-checks) on the files I added or modified and fixed the errors.
- [+ ] I have added the newly added features to **News.rst** (if applicable)


DominiqueMakowski:
Thanks :) It seems like it throws an error when a short signal is passed. Should we try to adjust nperseg or the overlap so that it doesn't fail in that case? If not,skip when shorter than some minimal length?

```python
import neurokit2 as nk

eda = nk.data('bio_resting_8min_100hz')['EDA']

indexes_posada = nk.eda_sympathetic(eda[0:2000], sampling_rate=100, method='posada', show=True)
```
```
ValueError: noverlap must be less than nperseg.
```

GanshengT:
> Thanks :) It seems like it throws an error when a short signal is passed. Should we try to adjust nperseg or the overlap so that it doesn't fail in that case? If not,skip when shorter than some minimal length?
> 
> ```python
> import neurokit2 as nk
> 
> eda = nk.data('bio_resting_8min_100hz')['EDA']
> 
> indexes_posada = nk.eda_sympathetic(eda[0:2000], sampling_rate=100, method='posada', show=True)
> ```
> 
>  ```
> ValueError: noverlap must be less than nperseg.
> ```

The algorithm is proposed for data that is more than 60 seconds. I suggest that we skip or maybe return a warning "Input signal should be more than 60 seconds when using posada method".

DominiqueMakowski:
Seems like there is bug in the nabian method that makes the docstring examples fail:

```
File D:\a\NeuroKit\NeuroKit\neurokit2\eda\eda_findpeaks.py:373, in _eda_findpeaks_nabian2018(eda_phasic)
    368 window = eda_phasic[i:j]
    369 # The amplitude of the SCR is obtained by finding the maximum value
    370 # between these two zero-crossings and calculating the difference
    371 # between the initial zero crossing and the maximum value.
    372 # amplitude defined in neurokit2
--> 373 amp = np.max(window)
    375 # Detected SCRs with amplitudes less than 10% of max SCR amplitude will be eliminated
    376 # we append the first SCR
    377 if len(amps_list) == 0:
    378     # be careful, if two peaks have the same amplitude, np.where will return a list

File <__array_function__ internals>:200, in amax(*args, **kwargs)

File C:\hostedtoolcache\windows\Python\3.10.10\x64\lib\site-packages\numpy\core\fromnumeric.py:2820, in amax(a, axis, out, keepdims, initial, where)
   2703 @array_function_dispatch(_amax_dispatcher)
   2704 def amax(a, axis=None, out=None, keepdims=np._NoValue, initial=np._NoValue,
   2705          where=np._NoValue):
   2706     """
   2707     Return the maximum of an array or maximum along an axis.
   2708 
   (...)
   2818     5
   2819     """
-> 2820     return _wrapreduction(a, np.maximum, 'max', axis, None, out,
   2821                           keepdims=keepdims, initial=initial, where=where)

File C:\hostedtoolcache\windows\Python\3.10.10\x64\lib\site-packages\numpy\core\fromnumeric.py:86, in _wrapreduction(obj, ufunc, method, axis, dtype, out, **kwargs)
     83         else:
     84             return reduction(axis=axis, out=out, **passkwargs)
---> 86 return ufunc.reduce(obj, axis, dtype, out, **passkwargs)

ValueError: zero-size array to reduction operation maximum which has no identity
```

GanshengT:
Could you tell me which example is failed?
I tried to run the following lines, no error is reported
eda_signal = nk.eda_simulate(duration=30, scr_number=5, drift=0.1, noise=0)
eda_cleaned = nk.eda_clean(eda_signal)
eda = nk.eda_phasic(eda_cleaned)
eda_phasic = eda["EDA_Phasic"].values
nabian2018 = nk.eda_findpeaks(eda_phasic, method="nabian2018")

DominiqueMakowski:
Changing max to nanmax seemed to have fixed the issue 🤷‍♂️ 

I've also taken the liberty of adding you as a contributor, do let me know if it's okay

DominiqueMakowski:
Actually this still does fail:

```python
      import neurokit2 as nk

      # Get phasic component
      eda_signal = nk.eda_simulate(duration=30, scr_number=5, drift=0.1, noise=0, sampling_rate=100)
      eda_cleaned = nk.eda_clean(eda_signal, sampling_rate=100)
      eda = nk.eda_phasic(eda_cleaned, sampling_rate=100)
      eda_phasic = eda["EDA_Phasic"].values
      _, nabian2018 = nk.eda_peaks(eda_phasic, sampling_rate=100, method="nabian2018")
```

Reason is that the windows are empty (`[]`), not sure how to best fix it

GanshengT:
The problem happens because a negative crossing happens before the first positive crossing. In this case, delete the first negative crossing would work.

DominiqueMakowski:
Could you add that or show me how to add?

DominiqueMakowski:
All good it seems, let me know if it's good to merge :)

GanshengT:
I agree to merge.

DominiqueMakowski:
Thanks a lot for diving into these functions, much appreciated

PR #755, opened 2022-11-19
[Feature] add rsp_methods (basis for rsp reports)

rostro36:
# Description

This PR is based on the discussion of [this thread](https://github.com/neuropsychology/NeuroKit/pull/750) and builds on my previous PR #754 . Let me know if I should point the PR to a different branch.

The goal here is to provide rsp_methods and follows the template made by @danibene  (thank you!) in ppg_methods, which was also refactored.

I likely won't have time for rsp_report in November, so somebody else might take over or I do it in December.
# Proposed Changes

- Creation of the rsp_methods function based on ppg_methods
- Refactor of ppg_methods and report.py

# Checklist

Here are some things to check before creating the PR. If you encounter any issues, do let us know :)

- [x] I have read the [CONTRIBUTING](https://github.com/neuropsychology/NeuroKit/blob/master/.github/CONTRIBUTING.rst#structure-and-code) file.
- [x] My PR is targetted at the **dev branch** (and not towards the master branch).
- [x] I ran the [CODE CHECKS](https://github.com/neuropsychology/NeuroKit/blob/master/.github/CONTRIBUTING.rst#run-code-checks) on the files I added or modified and fixed the errors.
- [x] I have added the newly added features to **News.rst** (if applicable)

PR #785, opened 2023-02-20
[Feature] add respiration report with interactive plot

danibene:
# Description

This PR builds on the `rsp_methods()` function added in https://github.com/neuropsychology/NeuroKit/pull/755 so that an interactive HTML report can be generated describing RSP processing as described in https://github.com/neuropsychology/NeuroKit/issues/665

# Proposed Changes

1. I changed the `rsp_plot()` function so that it optionally generates an interactive plot with plotly
2. I added an `rsp_report()` function similar to the existing `ppg_report()` function from https://github.com/neuropsychology/NeuroKit/pull/707 (there is a lot of repetition - I wonder if separate functions are really necessary and if this should rather be done with a general function in `misc`?)
3. I changed the `rsp_process()` function so that it optionally generates a report

# Checklist


danibene:
I think the reason the tests are failing is because of an issue with the `spicy` peak finding method, addressed here https://github.com/neuropsychology/NeuroKit/pull/786

DominiqueMakowski:
I am so grateful to have you on board in this project @danibene 🙏 🙌

DominiqueMakowski:
Let me know if you need help or review!

danibene:
@DominiqueMakowski yes review would be appreciated, in particular your thoughts on this: there is a lot of repetition - I wonder if separate functions are really necessary and if this should rather be done with a general function in misc?

DominiqueMakowski:
Well yeah, I think you're right, if there's a lot of similar plotting, we could add internals of type `_report_plot_signal()` in the report.py in misc and to avoid having redundant code

danibene:
@DominiqueMakowski I think it's ready now, I refactored so that `ppg_report()` and `rsp_report()` are now just replaced with a generalized function in the report.py in misc. I also returned figures regardless of whether `static=True` in `ppg_plot()` and `rsp_plot()` for consistency

DominiqueMakowski:
Impressive! I gave it a try it works, but it makes me wonder about this static argument, it would actually be nice to have a `static` arg in `create_report()` to allow people to have static figures (would be useful to convert then to pdf for non-html-friendly sharing). Do you think it is doable the way it is set up to allow create_report() to include either the static or the plotly versions? 

Then, we could document in the description of the `report` arg in the process functions that they can generate either by passing static=True as kwargs

danibene:
> It would actually be nice to have a `static` arg in `create_report()` to allow people to have static figures (would be useful to convert then to pdf for non-html-friendly sharing). Do you think it is doable the way it is set up to allow create_report() to include either the static or the plotly versions?

I changed `create_report()` so that either a matplotlib or plotly figure can be passed. 
> 
> Then, we could document in the description of the `report` arg in the process functions that they can generate either by passing static=True as kwargs
So here I'm not sure how to implement this, since the kwargs of the process functions are already used for the `_methods()` functions. I guess I could write a function so that any kwargs that are not used by the underlying processing functions are then passed to the plotting functions? But maybe that would be overly complicated?

DominiqueMakowski:
I see, perhaps we can leave it as is for now, and someday maybe we can do that if the extension is ".pdf" (or .word), it saves a static plot. But that would add some dependencies and logic so it's better to keep with one output format for now while we implement these reports :)

danibene:
> I see, perhaps we can leave it as is for now, and someday maybe we can do that if the extension is ".pdf" (or .word), it saves a static plot. But that would add some dependencies and logic so it's better to keep with one output format for now while we implement these reports :) 

That makes sense to me! Feel free to merge if you think it's ready now

PR #791, opened 2023-02-23
[Fix] Function used in rsp_methods

danibene:
# Description
I noticed that `rsp_methods()` uses the default arguments of `rsp_findpeaks()` instead of `rsp_peaks()`. `rsp_peaks()` is the function used in `rsp_process()` so I think we should use that. I also noticed that I left the plots generated in the report static here https://github.com/neuropsychology/NeuroKit/pull/785, so I fixed that. 

# Proposed Changes
1. I changed the `rsp_methods()` function to use `rsp_peaks()`
2. I changed the `rsp_process()` and `ppg_process()` functions to generate reports with interactive plots

# Checklist


DominiqueMakowski:
looks good, feel free to merge

PR #792, opened 2023-02-23
[Feature] add EDA report with interactive plot

danibene:
# Description

This PR aims to add the feature of generating an interactive HTML report describing EDA processing as described in https://github.com/neuropsychology/NeuroKit/issues/665

# Proposed Changes

1. I changed the `eda_plot()` function so that it optionally generates an interactive plot with plotly
2. I added a `eda_methods()` function to pass kwargs to the appropriate underlying processing functions. There still needs to be more done to fully describe all the methods (I wasn't sure which references were appropriate for `eda_phasic()` methods, for example), but this should still work to generate a report 
3. I changed the `eda_process()` function so that it optionally generates a report

# Checklist


danibene:
@DominiqueMakowski regarding your comment: https://github.com/neuropsychology/NeuroKit/issues/794#issuecomment-1445413686

Currently, cleaning is skipped in the `eda_process()` and `ppg_process()` functions, but I can also modify them so that it's done in the cleaning functions instead. I don't have any opinion on which one is better

DominiqueMakowski:
>  I don't have any opinion on which one is better

Perhaps better in the cleaning functions, unless it adds maintenance burden: this way people can also have a "skip step" option when they build their own custom processing pipelines? + it doesn't complexity the _process() which should remain very simple for people to see what it does. what do you think?

danibene:
> > I don't have any opinion on which one is better
> 
> Perhaps better in the cleaning functions, unless it adds maintenance burden: this way people can also have a "skip step" option when they build their own custom processing pipelines? + it doesn't complexity the _process() which should remain very simple for people to see what it does. what do you think?

makes sense to me, I updated the cleaning functions now

DominiqueMakowski (approved):

Issue #781, opened 2023-02-15
Offering help to investigate claim that cvxEDA is broken


lciti:
I am the developer of the cvxEDA code used in NeuroKit.

I would like to offer my help to investigate what sort of issues cvxEDA is giving and to address them, with reference to the comment below in the code:
https://github.com/neuropsychology/NeuroKit/blob/5cd089e9d82e6f7fb98d2028e69840aee08f8ab3/neurokit2/eda/eda_phasic.py#L17

There are two related issues in the cvxEDA repository (https://github.com/lciti/cvxEDA/issues/9 and https://github.com/lciti/cvxEDA/issues/10). The error reported there was raised by the cvxopt library with message "ValueError: Rank(A) < p or Rank([P; A; G]) < n". I have downloaded data that was kindly provided by the user who reported the error and installed the same version of the libraries but I could not reproduce the error under any conditions.

Because cvxEDA depends on cvxopt, which in turn depends on cholmod, which in turn depends on a BLAS/LAPACK implementation, it's very hard for me to reproduce the error and find a fix, unless I have the exact same data and libraries. I would be grateful if any person who receives this error could please export the exact list of packages with `conda env export > env.yaml` and attach it in reply to this issue, together with the data and platform. I am only interested in "reasonable" data and values of the parameters (e.g. the data cannot be just a handful of samples long, the sampling frequency cannot be 0.2 Hz or 500 MHz, because in these cases an error is reasonable).

There is the possibility that the error only happened with a specific platform and version of some of the libraries and that the underlying error in those libraries has then been fixed. In this case, I propose we monitor how cvxEDA is doing now and if it is not giving that error any more, we assume the problem has disappeared.

Thanks!

DominiqueMakowski:
Hi @lciti thanks for chiming in here. I was myself a big user of cvxEDA during my PhD, so I am really grateful and happy to see that it's being worked on ☺️ 

I have updated the latest cvxopt and here is a reproducible code snippet that errors for me: 

```python
import neurokit2 as nk

# Simulate EDA signal
eda_signal = nk.eda_simulate(duration=30, scr_number=5, drift=0.1)

nk.signal_plot(eda_signal)
```
![image](https://user-images.githubusercontent.com/8875533/219086431-20b85513-bee6-45eb-88ed-eae7989d2f52.png)

```python
cvxEDA = nk.eda_phasic(nk.standardize(eda_signal), method='cvxeda')
```
```
---------------------------------------------------------------------------
[...]

PR #793, opened 2023-02-25
[Improvement] Fixes #784 and use random generator objects rather than np.random global RNG

### LINKED ISSUE #784: nk.eda_simulate non reproducible even when random_state is set to a constant
**Describe the bug**
`nk.eda_simulate` generates random signals even when `random_state` is set to a constant

**To Reproduce**
```python
>>> eda_signal = nk.eda_simulate(duration=30, scr_number=5, drift=0.1, random_state=1234)
>>> print(eda_signal)
[0.98571118 0.98526776 0.98387649 ... 3.9812714  3.9785683  3.97764784]
>>> eda_signal = nk.eda_simulate(duration=30, scr_number=5, drift=0.1, random_state=1234)
>>> print(eda_signal)
[0.99979647 0.99998727 1.00034615 ... 4.05665343 4.05670009 4.05676347]
```

**Expected behaviour**
Setting `random_state` should make `nk.eda_simulate`  always return the same signal.
The problem is that `nk.eda_simulate` calls `signal_distort` (line [89](https://github.com/neuropsychology/NeuroKit/blob/8c564938c5fc2b4a377fd89bb14ac07a430983b9/neurokit2/eda/eda_simulate.py#L89)) without setting its `random_state`. This means that the argument is set to its default `None` and therefore line [106](https://github.com/neuropsychology/NeuroKit/blob/8c564938c5fc2b4a377fd89bb14ac07a430983b9/neurokit2/signal/signal_distort.py#L106) resets the random generator to a random state.
A solution is to call `signal_distort` with 'random_state=np.random.randint(np.iinfo(np.uint32).max)`.
An even better solution, moving forward, would be to use a see generator object and pass it around from one function to all its callees, as suggested for example [here](https://albertcthomas.github.io/good-practices-random-number-generators/).
A simple way to do this is to replace the
```python
np.random.seed(random_state)
```
at the beginning of all functions and replace it with:
```python
rng = np.random.default_rng(random_state)
```
and use `rng` throughout the function (e.g. we need to replace `np.random.normal(0, 5, size=1)` with `rng.normal(0, 5, size=1)`) and pass it to all callees as `random_state=rng`. We also need to remove `np.random.seed(None)` from the end of each function (since we are not touching the singleton numpy random generator, there's no need to reset it to a random state).

lciti:
# Description

This PR fixes #784 and make changes to the way random numbers are generated along the lines of what I describe in the discussion of #784.

# Proposed Changes

Modified the approach to random number generation (also solving the issue that the same seed was used both to generate a signal and its noise).
XXX_simulate functions now use an RNG object (RandomState or Generator) rather than the global singleton numpy.random one.
Their argument` random_state` can take several values, notably the following ones:
- When this is an integer, a RandomState object is created and used (less performant but more reproducible).
- When this is None, a Generator object is created and used (more performant but without reproducibility guarantees for different versions of numpy, it is also the recommended RNG for new code by the [numpy manual](https://numpy.org/doc/stable/reference/random/index.html)).
- When it's already a RandomState or Generator, this object is used to generate random values.

Some ways to generate random numbers had to be changed to use methods available both in RandomState and in Generator, e.g. `randn`->`standard_normal` (since `randn` is in RandomState but not in Generator, while `standard_normal` is in both).

Regarding the issue that the same seed was used both to generate a signal and its noise, now the XXX_simulate functions also take `random_state_distort`, which can take value "legacy" (to reproduce results up to nk 0.2.3) or "spawn" to generate noise independent of the signal. It can also take, e.g., a seed, which allows creating different simulated signals with the same "signal" and different random instances of "noise" added to it. This may be useful to determine whether a given processing algorithm (e.g. peak detection) produces similar results when fed the same signal corrupted by different random instances of noise.

*NOTE:* I haven't added any documentation or example of the extra functionality or extra argument yet. Before investing additional time, I want to make sure @DominiqueMakowski is happy with the changes.

# Checklist


lciti:
Failed the new test for k-means. It looks like the test suite uses a version of sk-learn that does not implement the [`n_init='auto'` argument](https://github.com/scikit-learn/scikit-learn/commit/0d923daf6ddbf1afc0915a145955cfff97505f51), which `nk` assumes available here: https://github.com/neuropsychology/NeuroKit/blob/8c564938c5fc2b4a377fd89bb14ac07a430983b9/neurokit2/stats/cluster.py#L190
It works fine on my machine with sk-learn 1.2.1.

tests/tests_rsp.py
```diff
@@ -31,6 +32,60 @@ def test_rsp_simulate():
...
+        (13579, "spawn"),
+        (13579, 24680),
+        (13579, None),
+        (np.random.RandomState(33), "spawn"),
+        (np.random.SeedSequence(33), "spawn"),
+        (np.random.Generator(np.random.Philox(33)), "spawn"),
+        (None, "spawn"),
+    ],
+)
+def test_ppg_simulate_all_rng_types(random_state, random_state_distort):
```
danibene, line 70: I think you meant rsp, not ppg

danibene:
> Failed the new test for k-means. It looks like the test suite uses a version of sk-learn that does not implement the [`n_init='auto'` argument](https://github.com/scikit-learn/scikit-learn/commit/0d923daf6ddbf1afc0915a145955cfff97505f51), which `nk` assumes available here: https://github.com/neuropsychology/NeuroKit/blob/8c564938c5fc2b4a377fd89bb14ac07a430983b9/neurokit2/stats/cluster.py#L190
> It works fine on my machine with sk-learn 1.2.1.

@DominiqueMakowski how would you suggest addressing this? Changing the version of sklearn used in the test suite/required by neurokit, or changing the code e.g. so that the number of clusters is provided as an integer if `n_init=auto` fails?

danibene (commented):
This generally looks good to me, I think all that remains to do is renaming the rsp test function from ppg, adding documentation, and addressing the failing cluster test

tests/tests_rsp.py
lciti: Yes, I think you are right.

neurokit2/misc/random.py
```diff
@@ -0,0 +1,57 @@
...
+    # If seed is already a random number generator return it as it is
+    if isinstance(seed, (np.random.Generator, np.random.RandomState)):
+        return seed
+    # If seed is something else, use the new Generator class
+    # Note: to initialise the new generator class with an integer seed, use, e.g.:
+    # check_rng(np.random.SeedSequence(123))
+    return np.random.default_rng(seed)
+
+
+def spawn_rng(rng, n_children=1):
```
DominiqueMakowski, line 22: This could probably be made an internal `_spawn_rng` and not exported as it's only used in get_children_rng afaik

DominiqueMakowski:
Thanks for that impressive PR! 

Maybe we could change the names, what do you think about `sanitize_randomseed()` (for check_rng) and `multiple_randomseeds()`?

neurokit2/misc/random.py
lciti: It is only used internally for now but I would leave it as public as spawning an rng is good practice whenever several RNGs are needed. For example, if one needs to simulate several signals (e.g. several patients in a study) one should not reuse the same seed or use seed, seed+1, seed+2... but rather 
```python
master_seed = 42
patients_RNGs = nk.misc.spawn_rng(master_seed, n_children=n_patients)
PPGs = []
for i in range(n_patients):
    PPGs.append(nk.ppg_simulate(..., random_state=patients_RNGs[i]))
```

lciti:
@DominiqueMakowski I agree the names are a bit strange. However I called them this way based on the names of virtually identical functions in other python projects. See for example [sklearn's `check_random_state`](https://scikit-learn.org/stable/modules/generated/sklearn.utils.check_random_state.html), [statsmodels's one](https://www.statsmodels.org/dev/generated/statsmodels.tools.rng_qrng.check_random_state.html) and [scipy's one](https://github.com/scipy/scipy/blob/0655c14c549f71200ed6a8a4938098270c4352dc/scipy/_lib/_util.py#L177), which do virtually the same thing. I decided to call it `check_rng` rather than `check_random_state` to make it closer to the new recommended numpy Generator API, however we could also call it `check_random_state` for consistency with these other projects. To be honest, I initially called it `get_rng` (or `get_random_state`) but then I found the sklearn's one and tried to be consistent. Anyway, I am open to reconsider the choice (as I am not 100% sure myself).
As for `spawn`, that's how the new numpy API will call this function, see [here](https://github.com/numpy/numpy/pull/23195). The function `spawn_rng` I implemented basically calls the numpy `spawn` method when available (only in the next release and only for `Generator`, not for `RandomState`) and does the same thing or something close in the other cases. In this case I think `spawn` is the right term (but could be `spawn_random_state`).

lciti:
@danibene , @DominiqueMakowski
It's ready to go on my side. Please let me know if you have any comments.

danibene:
This looks good to me!

DominiqueMakowski:
Thanks a lot @lciti for that impressive work :) merging now

lciti:
My pleasure! Many thanks to you for developing and maintaining such an impressive tool!
I hope we can fix that other bug too (https://github.com/neuropsychology/NeuroKit/issues/781)!

PR #797, opened 2023-02-27
[Feature] ecg_interpolatepeaks

the-visserd:
Hi,

long-time user & first-time contributor here. 

I made a feature that allows you to interpolate R-peaks trigonometrically. In the literature, you often see a linear interpolation of IBIs. However, this neglects the fact that the heart beat is a periodic signal. What I mean by the periodicity of the signal is best shown in a plot:

output

Here, y=1 stands for an R-peak, y=0 for the center point between two R-peaks. Note that this feature is a mere interpolation based on a cosine. The frequency of a period shifts accordingly to the interbeat interval. It is not a transformation, like FFT. I hope this function can be helpful in the analysis of IBIs as a time series. 

Cheers

_Edit: Fixed a bug & added a test file. Feel free to decide if this is a function you want to have in neurokit2 :)_

# Checklist

### DIFF
--- a/neurokit2/ecg/ecg_interpolatepeaks.py
+++ b/neurokit2/ecg/ecg_interpolatepeaks.py
@@ -1,0 +1,60 @@
+from math import cos, pi
+import numpy as np
+import pandas as pd
+
+
+def ecg_interpolatepeaks(ecg_peaks):
+    """**Interpolate R-peaks in an ECG signal**
+
+    Trigonometrically interpolates between R-peaks. This approach considers the intrinsic 
+    periodical property of heartbeats, instead of a mere linear interpolation between IBIs.
+    As a period can only be calculated with a known start and ending, the entries before the
+    first R-peak and the entries after the last R-peak will be np.NaNs. The length of the 
+    original time series is preserved.
+
+    Parameters
+    ----------
+    ecg_peaks : tuple
+        Takes the output of ecg_peaks().
+
+    Returns
+    ----------
+    rpeaks_cip : pd.Series
+        A cosine interpolation of the R-peaks between the values 1 and 0 for y, where 1 stands
+        for an R-peak and 0 stands for the center point between two subsequent R-peaks.
+    """
+    def cosine(x, x1, x2):
+        # Returns cosine value (y) [0,1] for given time step (x) in one period (T) based on
+        # the interval between x1 and x2.
+        # Args:
+        #    x (float): time step
+        #    x1 (float): period start (y=1)
+        #    x2 (float): period end (y=1)
+        # Returns:
+        #    val (float): cosine value [0,1]
+
+        T = x2-x1
+        val = 0.5+0.5*cos(2.0 * pi * (x - x1) / T)
+        return val
+
+    length = len(ecg_peaks[0])
+    rpeaks_idx = ecg_peaks[1]["ECG_R_Peaks"]
+    num_rpeaks = len(rpeaks_idx)
+
+    # Interpolation
+    rpeaks_cip = [np.nan]*(rpeaks_idx[0])
+    for i in range(num_rpeaks):
+        if i == num_rpeaks-1:
+            break
+        x1 = rpeaks_idx[i]
+        x2 = rpeaks_idx[i+1]
+        T = x2-x1
+        def f(x): return cosine(x, x1, x2)
+        # span interval with corresponding number of time steps
+        x = np.linspace(x1, x2, T)
+        y = [f(x) for x in x]
+        rpeaks_cip = rpeaks_cip + y
+    rpeaks_cip = rpeaks_cip + [np.nan]*(length-rpeaks_idx[-1])
+    # keep in mind: there are nans at the start and end
+
+    return pd.Series(rpeaks_cip)
--- a/neurokit2/ecg/ecg_interpolatepeaks.py
+++ b/neurokit2/ecg/ecg_interpolatepeaks.py
@@ -9,8 +9,8 @@ def ecg_interpolatepeaks(ecg_peaks):
     Trigonometrically interpolates between R-peaks. This approach considers the intrinsic 
     periodical property of heartbeats, instead of a mere linear interpolation between IBIs.
     As a period can only be calculated with a known start and ending, the entries before the
-    first R-peak and the entries after the last R-peak will be np.NaNs. The length of the 
-    original time series is preserved.
+    first R-peak and the entries after the last R-peak will be NaN. The length of the orginal
+    time series is preserved.
 
     Parameters
     ----------
@@ -42,6 +42,7 @@ def ecg_interpolatepeaks(ecg_peaks):
     num_rpeaks = len(rpeaks_idx)
 
     # Interpolation
+    # add NaNs for the time steps before the first R-peak
     rpeaks_cip = [np.nan]*(rpeaks_idx[0])
     for i in range(num_rpeaks):
         if i == num_rpeaks-1:
@@ -50,11 +51,14 @@ def ecg_interpolatepeaks(ecg_peaks):
         x2 = rpeaks_idx[i+1]
         T = x2-x1
         def f(x): return cosine(x, x1, x2)
-        # span interval with corresponding number of time steps
-        x = np.linspace(x1, x2, T)
+        # span interval with corresponding number of time steps (see next comment why T+1)
+        x = np.linspace(x1, x2, T+1)
         y = [f(x) for x in x]
-        rpeaks_cip = rpeaks_cip + y
-    rpeaks_cip = rpeaks_cip + [np.nan]*(length-rpeaks_idx[-1])
-    # keep in mind: there are nans at the start and end
+        # period must not end with y=1 as it also starts with a 1 and we want only one R-peak
+        rpeaks_cip = rpeaks_cip + y[:-1]
+    # the last period ends with a 1
+    rpeaks_cip.append(1)
+    # add NaNs for the time steps after the last R-peak
+    rpeaks_cip = rpeaks_cip + [np.nan]*(length-rpeaks_idx[-1]-1)
 
     return pd.Series(rpeaks_cip)
--- a/tests/tests_ecg_interpolatepeaks.py
+++ b/tests/tests_ecg_interpolatepeaks.py
@@ -1,0 +1,34 @@
+import os.path
+import numpy as np
+import pandas as pd
+import neurokit2 as nk
+from neurokit2.ecg.ecg_interpolatepeaks import ecg_interpolatepeaks
+
+def test_ecg_findpeaks():
+    # simulate data
+    ecg_clean_peaks = nk.ecg_peaks(
+        nk.ecg_clean(
+            nk.ecg_simulate(
+                duration=10,
+                sampling_rate = 1000,
+                heart_rate=100, 
+                random_state=42)))
+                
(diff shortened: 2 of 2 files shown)


DominiqueMakowski:
hi @the-visserd and thanks for your contribution! 

Looks interesting, but I am having trouble understanding 1) what does it exactly do: in the figure you attached, is the orange one the result of the interpolatepeaks function? And 2) why would one want to do that? Like what would be typical usecases?

Thanks!

the-visserd:
Hey Dominique,

Thanks for the welcome. Correct, the orange line is the result of the trigonometric interpolation.

Pertaining to your second question - Many cognitive neuroscientists look at IBI values instead of ECG signals, because ECG signals are quite complex (obviously). If you want to have a time series of IBI values, many researchers rely on linear interpolations. From [Allen et al. (2007)](http://apsychoserver.psych.arizona.edu/JJBAReprints/PSYC501A/Readings/Allen_Chambers_Towers_Bio_Psychology_2007.pdf), p258: "An [array of IBI values] in not, strictly speaking, a time series, as the data occur at uneven intervals, provided that there is variability in cardiac chronotropy, the very phenomenon of interest! An IBI series can be converted to a time series by interpolating data points at a fixed sampling rate. [We recommend] a 10 Hz sampling rate with linear interpolation"

Unfortunately, this leads to bias (you lose information) - for multivariate time series even to statistical artefacts (you unify the periodic qualities). A trigonometric interpolation is the only way to correctly interpolate a periodic signal like R-peaks as a time series. While trigonometric interpolations are standard in engineering, in psychology they haven't really caught up yet.

If you have the feeling that this function does not fit well into neurokit2, that's okay with me (it has been a good learning experience for me, as this is my first PR).

DominiqueMakowski:
So if I understand, it is a new interpolation method to compute the heart "rate" (or period).

In NK, we have functions like `ecg_rate()`, `rsp_rate()` that are basically copies of a more generic function called `signal_rate()` that takes "peaks" (or beat indices) and returns the rate, interpolated or not. By default, ecg_rate() uses a monotonic-cubic interpolation method. 

Based on what you are saying, it seems like your function would be better added as a new method inside [signal_interpolate()](https://github.com/neuropsychology/NeuroKit/blob/master/neurokit2/signal/signal_interpolate.py#L7) (which is used by signal_rate()). 

What I would suggest is to add your interpolation method in that file, as `_signal_interpolate_trigonometric()`, and then add another `elif` step in the main `signal_interpolate()` function that calls your function if `method=="trigonometric"`. From there it should be possible for users to select this interpolation method when processing their ECG, but first I'll be curious to see how it compares with other interpolation methods.

Let me know if you have any questions!

Looping in @danibene

danibene:
> Based on what you are saying, it seems like your function would be better added as a new method inside [signal_interpolate()](https://github.com/neuropsychology/NeuroKit/blob/master/neurokit2/signal/signal_interpolate.py#L7) (which is used by signal_rate()). 
> 
> What I would suggest is to add your interpolation method in that file, as `_signal_interpolate_trigonometric()`, and then add another `elif` step in the main `signal_interpolate()` function that calls your function if `method=="trigonometric"`. From there it should be possible for users to select this interpolation method when processing their ECG, but first I'll be curious to see how it compares with other interpolation methods.

Yes, I agree, adding to `signal_interpolate()` and comparing to the interpolation methods that are already available beyond `linear` makes sense to me.

the-visserd:
Thanks for your guidance, Dominique and danibene - I really appreciate it.

Sorry for my late reply. I've had the flu and will get back to you here by the end of this week at the latest.

danibene:
No rush @the-visserd hope you feel better soon!

the-visserd:
So, I tried to add my function to signal_interpolate() but it didn't work out. As far as I can see, this stems from the fact that the functions ecg_interpolatepeaks() and ecg_rate() are in different domains. While ecg_interpolatepeaks() is in the time domain, ecg_rate is in the period domain (meaning we have individual heart beats on the x-axis). In my opinion, both are fine approaches. 

However, when we have multivariate data and we don't want to work with highly complex ECG data, we'd need an interpolation with equidistant spacing in the time domain between each data point. That is where my function could come into play (ecg_interpolatepeaks() gives a continuous signal for a discrete R-peak signal; see figure below). 

I am curious what your thoughts are on this. If you think this is too niche for nk2, that's fine by me.


grafik


***
Code for the data & figure above:
````python

from math import cos, pi
import numpy as np
import pandas as pd


def ecg_interpolatepeaks(ecg_peaks):
    """**Interpolate R-peaks in an ECG signal**

    Trigonometrically interpolates between R-peaks. This approach considers the intrinsic 
    periodical property of heartbeats, instead of a mere linear interpolation between IBIs.
    As a period can only be calculated with a known start and ending, the entries before the
    first R-peak and the entries after the last R-peak will be NaN. The length of the orginal
    time series is preserved.

    Parameters
    ----------
    ecg_peaks : tuple
        Takes the output of ecg_peaks().

    Returns
    ----------
    rpeaks_cip : pd.Series
        A cosine interpolation of the R-peaks between the values 1 and 0 for y, where 1 stands
        for an R-peak and 0 stands for the center point between two subsequent R-peaks.
    """
    def cosine(x, x1, x2):
        # Returns cosine value (y) [0,1] for given time step (x) in one period (T) based on
        # the interval between x1 and x2.
        # Args:
        #    x (float): time step
        #    x1 (float): period start (y=1)
        #    x2 (float): period end (y=1)
        # Returns:
        #    val (float): cosine value [0,1]

        T = x2-x1
        val = 0.5+0.5*cos(2.0 * pi * (x - x1) / T)
        return val

    length = len(ecg_peaks[0])
    rpeaks_idx = ecg_peaks[1]["ECG_R_Peaks"]
    num_rpeaks = len(rpeaks_idx)

    # Interpolation
    # add NaNs for the time steps before the first R-peak
    rpeaks_cip = [np.nan]*(rpeaks_idx[0])
    for i in range(num_rpeaks):
        if i == num_rpeaks-1:
            break
        x1 = rpeaks_idx[i]
        x2 = rpeaks_idx[i+1]
        T = x2-x1
        def f(x): return cosine(x, x1, x2)
        # span interval with corresponding number of time steps (see next comment why T+1)
        x = np.linspace(x1, x2, T+1)
        y = [f(x) for x in x]
        # period must not end with y=1 as it also starts with a 1 and we want only one R-peak
        rpeaks_cip = rpeaks_cip + y[:-1]
    # the last period ends with a 1
    rpeaks_cip.append(1)
    # add NaNs for the time steps after the last R-peak
    rpeaks_cip = rpeaks_cip + [np.nan]*(length-rpeaks_idx[-1]-1)

    return pd.Series(rpeaks_cip)
````
````python
import numpy as np
import neurokit2 as nk

# Generate singals
ecg = nk.ecg_simulate(heart_rate=80)
ecg_peaks = nk.ecg_peaks(ecg)
ecg_peaks_interpolated = ecg_interpolatepeaks(ecg_peaks)
ecg_rate = nk.ecg_rate(ecg_peaks)

# Plot signals
fig, axs = plt.subplots(4,1, figsize=(7,7), dpi=100)
axs[0].set_title('ECG signal from ecg_simulate()')
axs[0].plot(ecg)
axs[1].set_title('R-peaks from ecg_peaks()')
axs[1].plot(ecg_peaks[0])
axs[2].set_title('Interpolated R-peaks from ecg_interpolatepeaks()')
axs[2].plot(ecg_peaks_interpolated)
axs[2].set_xlabel('Time [ms]')
[axs[i].set_xlim([0,len(ecg)]) for i in range(3)]
axs[3].set_title('Signal rate (per minute) for periods from ecg_rate()')
axs[3].plot(ecg_rate)
axs[3].set_xticks(list(range(len(ecg_rate))))
axs[3].set_xlabel('Period [idx]')
fig.tight_layout()
````

danibene:
@the-visserd maybe you could separately transform the ECG signal according to your definition, where 1 stands for an R-peak and 0 stands for the center point between two subsequent R-peaks, and make the rest of the points in the signal NaN, then apply `signal_interpolate()`? 

Regarding whether it makes sense to add this in the first place, it would be good to understand the use cases of interpolating R-Peaks rather than interbeat intervals (if I understand correctly the literature you mentioned before interpolated interbeat intervals at 10 Hz). Do you know of any literature interpolating the R-Peaks in this way instead of the interbeat interval signal? Or have you found any benefit of doing this yourself?

the-visserd:
@danibene I'm having difficulties visualizing how my way of interpolating *binary discrete* data could be incorporated into `signal_interpolate()` which is based on *non-binary continuous* data (see figure below for the examples of the function - how would you draw a cosine interpolation in this?)

grafik

At the moment, I'm thinking that my feature would fit best into the function `ecg_peak()`. The default output would be the binary discrete time series as is, and there could be the option for a continuous interpolation of the R-peaks.

Here are my top reasons why I think this feature is important:
- Axioms: Heartbeats are periodic signals and should be handled as such in time series. A linear perspective seemingly might be easier to work with, but will ultimately lead to inaccuracies.
- Representation: Linearly interpolating IBIs leads to the unit `time / heartbeat` on the y-axis and `time` on the x-axis, which in combination is strange to say the least.
- Time series: For complex non-stationary data, continuous values are easier to handle than binary data, as you'd never know at which exact time step the signal will be 1 out of hundreds of consecutive 0s. A cosine interpolation of R-peaks solves this problem by delivering continuous values between two 1s.

For more information and further reading on this, you can check out the last paragraph of https://www.kubios.com/hrv-time-series/ (some studies suggest polynomial fitting, but I think my code solution is more practical).

the-visserd:
Fair point - but let me ask you a question. What would you do if you want to analyze the occurrence of R-peaks in a time series where an ECG gives too much information and a binary time series too little?

Also, the idea behind the linear interpolation of IBIs is to bring the values back to the time domain (with strange units besides) - with a cosine interpolation you can stay in the time domain and you don't have to do strange transformations from an epoch domain (IBI) back to a time domain (IBI/time).

danibene:
What I was curious about was whether we could just transform the ECG R-Peaks to the format you'd like (1s being peaks and 0s being the midpoint between peaks) and use one of the existing interpolation functions, e.g. `monotone_cubic`: https://gist.github.com/danibene/14df8d8facc9dac3796d0ba276409eb7

> Fair point - but let me ask you a question. What would you do if you want to analyze the occurrence of R-peaks in a time series where an ECG gives too much information and a binary time series too little?
> 
> Also, the idea behind the linear interpolation of IBIs is to bring the values back to the time domain (with strange units besides) - with a cosine interpolation you can stay in the time domain and you don't have to do transformations from an epoch domain (IBI/epoch) back to a time domain ((pseudo-)IBI/time).

From what I have read the transformation to interbeat intervals (regardless of what type of interpolation is used) is pretty standard for HRV analysis, and this is what I would normally do if I am not interested in the other features of the ECG.

the-visserd:
Besides the first and last episode this looks good! So the existing interpolation methods may be already satisfactory. 

Linearly interpolated interbeat intervals being a standard is not a good argument for its validity. In my opinion, an "instantaneous heart rate doesn't make much sense, but nor does "half a heart beat", as Dominique pointed out. Nevertheless, with the latter the information about the number of heart beats and their exact moment of occurence remains intact (hopefully you don't have 10 heartbeats per second for an interpolated interbeat intervals at 10 Hz). 

My hope is that an interpolation in the time domain catches up one day - we'll see.

Issue #799, opened 2023-03-06
Unvevenly spaced Lomb, new feature issue.


RichardBJ:
**Describe the bug**
Running HRV analysis (frequency) with the Lomb method, including unevenly spaced data. I fully appreciate it could be my inputs, but in that case perhaps the doc string could be spelt out more.

 **error log** (the message related to the error).
I've tried many different attempts and so different errors, but for example:
`ValueError: NeuroKit error: _hrv_sanitize_input(): The peak indices passed were detected as non-consecutive. You might have passed RR intervals instead of peaks. If so, convert RRIs into peaks using nk.intervals_to_peaks().`
**To Reproduce**
Some data:
[rrs.xlsx](https://github.com/neuropsychology/NeuroKit/files/10903319/rrs.xlsx)
So this works:

```
from astropy.timeseries import LombScargle
import matplotlib.pyplot as plt  
import pandas as pd
import neurokit2 as mykit
HRs=pd.read_excel("rrs.xlsx")
frequency, power = LombScargle(HRs["Time"], HRs["RR"], normalization="psd").autopower(
    minimum_frequency=0.1, maximum_frequency=3)
plt.plot(frequency, power) 

```
BUT this does not:
```
hrv_freq = mykit.hrv_frequency(HRs["RR"],
        psd_method="lombscargle", show=True, silent=False, normalize=True,
       interpolation_rate=None,t=HRs["Time"])

```
**Expected behaviour**
would expect a periodograph produced, as with astropy, but prettier.
So to confirm; works with a Welch and with astropy's lomb, but not my unevenly spaced lomb.  Could be the way I am calling the function, but I have tried several different methods.

**System Specifications**
iMac: Python 3.9?

danibene:
Hi @RichardBJ thanks for this! I submitted a PR (https://github.com/neuropsychology/NeuroKit/pull/800) and tested the new version of the code with your data:
https://gist.github.com/danibene/abbc76fdf6d3b447a216b1d69f791325

> Running HRV analysis (frequency) with the Lomb method, including unevenly spaced data. I fully appreciate it could be my inputs, but in that case perhaps the doc string could be spelt out more.

Yes, to have the RRI intervals and their timestamps as input, the first argument needed to be a dictionary containing both the R-R intervals and their timestamps with the keys `RRI` and `RRI_Time`, rather than using `t` as a separate argument

[...]

PR #800, opened 2023-03-07
[Fix] frequency-domain HRV with lombscargle given unevenly spaced data

danibene:
# Description

This PR aims to fix https://github.com/neuropsychology/NeuroKit/issues/799 to plot frequency-domain HRV analysis with the `lombscargle` PSD method given unevenly spaced data 

# Proposed Changes

1. I changed the `_hrv_frequency_show()` and `signal_psd()` so that when `sampling_rate=None`, the sampling rate is estimated using the median difference between timestamps 
_side note: this choice was somewhat arbitrary, I could have also used the mean or mode difference, I thought about adding a function along the lines of `_estimate_sampling_rate(t, method="mode")`, but wasn't sure where to put it_

# Checklist

### DIFF
--- a/neurokit2/hrv/hrv_frequency.py
+++ b/neurokit2/hrv/hrv_frequency.py
@@ -265,12 +265,19 @@ def _hrv_frequency_show(
         __, ax = plt.subplots()
 
     frequency_band = [ulf, vlf, lf, hf, vhf]
+
+    # Compute sampling rate for plot windows
+    if sampling_rate is None:
+        med_sampling_rate = np.median(np.diff(t))
+    else:
+        med_sampling_rate = sampling_rate
+
     for i in range(len(frequency_band)):  # pylint: disable=C0200
         min_frequency = frequency_band[i][0]
         if min_frequency == 0:
             min_frequency = 0.001  # sanitize lowest frequency
 
-        window_length = int((2 / min_frequency) * sampling_rate)
+        window_length = int((2 / min_frequency) * med_sampling_rate)
         if window_length <= len(rri) / 2:
             break
 
--- a/neurokit2/hrv/hrv_frequency.py
+++ b/neurokit2/hrv/hrv_frequency.py
@@ -268,7 +268,7 @@ def _hrv_frequency_show(
 
     # Compute sampling rate for plot windows
     if sampling_rate is None:
-        med_sampling_rate = np.median(np.diff(t))
+        med_sampling_rate = np.median(np.diff(t))  # This is just for visualization purposes (#800)
     else:
         med_sampling_rate = sampling_rate
 
--- a/neurokit2/signal/signal_psd.py
+++ b/neurokit2/signal/signal_psd.py
@@ -137,7 +137,10 @@ def signal_psd(
     # Sanitize min_frequency
     N = len(signal)
     if isinstance(min_frequency, str):
-        min_frequency = (2 * sampling_rate) / (N / 2)  # for high frequency resolution
+        if sampling_rate is None:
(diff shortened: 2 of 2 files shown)


RichardBJ:
Thanks for this, I have switched to this branch and now pretty closely reproduces a native astropy lomb (but with Neurokit's enhancements).  The sample_rate, surely if one is down this route (lomb, no interpolation, unevenly spaced), the sample_rate is unnecessary apart from the calculation of default min/max frequencies?  Is that right? In which case, I am not sure if it really matters which flavour of average is used. I imagine the vaguely fussy will set their own limits with the vlf=() options etc already?

danibene:
> Thanks for this, I have switched to this branch and now pretty closely reproduces a native astropy lomb (but with Neurokit's enhancements).  The sample_rate, surely if one is down this route (lomb, no interpolation, unevenly spaced), 

Glad to hear!

> the sample_rate is unnecessary apart from the calculation of default min/max frequencies?  Is that right? 

It's only used for generating windows for plotting and calculating the default min frequency (which can be otherwise set by the user), so that's true, the method for determining the sample rate when the data is unevenly spaced is not really crucial

DominiqueMakowski:
> It's only used for generating windows for plotting and calculating the default min frequency

Then what I would do is to take the median as currently done (good default choice I think) but then in the plot replace the values to make it explicit that the sampling rate info is missing? Basically I am worried that if it gives some values, people will assume it is correct and won't think of supplying the sampling rate... unless I didn't follow (I just moved in (finally) to a flat in the UK last week but I still don't have internet or a chair there, so can't be very productive for now - so Dani don't hesitate to take the lead on time-sensitive stuff ☺️)

danibene:
> > It's only used for generating windows for plotting and calculating the default min frequency
> 
> Then what I would do is to take the median as currently done (good default choice I think) but then in the plot replace the values to make it explicit that the sampling rate info is missing? Basically I am worried that if it gives some values, people will assume it is correct and won't think of supplying the sampling rate... unless I didn't follow

I don't think there are any misleading values displayed so I'll leave it as is for now

> (I just moved in (finally) to a flat in the UK last week but I still don't have internet or a chair there, so can't be very productive for now - so Dani don't hesitate to take the lead on time-sensitive stuff ☺️)

Ah wow congrats :) and good to know. I'll leave this PR open for you to review later since my branch can just be used in the meantime and it's not a critical fix imo

RichardBJ:
Best Wishes with the new flat!

> Basically I am worried that if it gives some values, people will assume it is correct and won't think of supplying the sampling rate

No I don't think it would be incorrect, it just might not be showing the range they wanted?

neurokit2/hrv/hrv_frequency.py
```diff
@@ -265,12 +265,19 @@ def _hrv_frequency_show(
         __, ax = plt.subplots()
 
     frequency_band = [ulf, vlf, lf, hf, vhf]
+
+    # Compute sampling rate for plot windows
+    if sampling_rate is None:
+        med_sampling_rate = np.median(np.diff(t))
```
DominiqueMakowski, line 271: Can we add a comment here to justify this decision, e.g.

```suggestion
        med_sampling_rate = np.median(np.diff(t))  # This is just for visualization purposes (see #800)
```

RichardBJ: OK thanks, I am sure you'll be correct! I don't have the code open right now!

danibene: I've now added a separate comment to `signal_psd()`, thank you :)

DominiqueMakowski (approved):

PR #810, opened 2023-03-29
[Docs] add facecolor argument

the-visserd:
Adds a white facecolor for all figures in the readme. At the moment, users of the dark github theme can't see text elements around a plot, because the facecolor is transparent. Note that the new figures still have to be generated with the next deployment.


DominiqueMakowski:
good idea, thanks!

PR #814, opened 2023-04-06
0.2.5

DominiqueMakowski:


danibene:
It seems the tests are failing because the newest version of sklearn wants us to use `valid_metrics()` instead of `valid_metrics`:

https://scikit-learn.org/stable/modules/generated/sklearn.neighbors.KDTree.html

DominiqueMakowski:
I suspect that some dependency creates some problems downstream, because tests are failing despite not having changed anything critical afaik

DominiqueMakowski:
Should we merge?

danibene:
> Should we merge?

Yes! I'll let you do the honors here

DominiqueMakowski:
Of course I had forgotten unpushed changes 😶‍🌫️ 
I added `signal_fillmissing()` to start adding a bit of control / flexibility in some cleaning pipelines (touched only PPG for now)

danibene:
@DominiqueMakowski I think we can merge with this! 

To be considered maybe in a future version if we want to rather adopt the peak detection functions to exclude periods of missing data, especially if these periods are longer (so that we don't have a distorted heart rate due to filled in data)

Issue #824, opened 2023-05-13
Failing test_eda_peaks when comparing to biosppy


danibene:
**Describe the bug**
The following test fails, I suspect due to the version of dependencies: https://github.com/neuropsychology/NeuroKit/actions/runs/4959364378/jobs/8873360950?pr=822

```
=================================== FAILURES ===================================
________________________________ test_eda_peaks ________________________________

    def test_eda_peaks():
        sampling_rate = 1000
        eda = nk.eda_simulate(
            duration=30 * 20,
            sampling_rate=sampling_rate,
            scr_number=6 * 20,
            noise=0,
            drift=0.01,
            random_state=42,
        )
        eda_phasic = nk.eda_phasic(nk.standardize(eda), method="highpass")["EDA_Phasic"].values
    
        signals, info = nk.eda_peaks(eda_phasic, method="gamboa[200](https://github.com/neuropsychology/NeuroKit/actions/runs/4959364378/jobs/8873360950?pr=822#step:5:201)8")
        onsets, peaks, amplitudes = biosppy.eda.basic_scr(eda_phasic, sampling_rate=1000)
        assert np.allclose((info["SCR_Peaks"] - peaks).mean(), 0, atol=1e-5)
    
        signals, info = nk.eda_peaks(eda_phasic, method="kim2004")
>       onsets, peaks, amplitudes = biosppy.eda.kbk_scr(eda_phasic, sampling_rate=1000)

tests/tests_eda.py:106: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

signal = array([-0.08155[226](https://github.com/neuropsychology/NeuroKit/actions/runs/4959364378/jobs/8873360950?pr=822#step:5:227), -0.0815639 , -0.08157554, ..., -0.00028455,
       -0.00025869, -0.000[232](https://github.com/neuropsychology/NeuroKit/actions/runs/4959364378/jobs/8873360950?pr=822#step:5:233)82])
sampling_rate = 1000, min_amplitude = 0.1

    def kbk_scr(signal=None, sampling_rate=1000., min_amplitude=0.1):
        """KBK method to extract Skin Conductivity Responses (SCR) from an
        EDA signal.
    
        Follows the approach by Kim *et al.* [KiBK04]_.
    
        Parameters
        ----------
        signal : array
            Input filterd EDA signal.
[...]

PR #825, opened 2023-05-13
[Fix] remove eda biosppy test

danibene:
# Description

Removes a failing test comparing an EDA method with the biosppy package. See https://github.com/neuropsychology/NeuroKit/issues/824 

# Proposed Changes

I changed the `test_eda_peaks()` function so that the `kim2004` method is not compared with biosppy.

# Checklist


danibene:
I'm going to try to fix the documentation check here but we can revert if we decide that doesn't make sense

DominiqueMakowski:
Let's merge then, I'll give it a look later

danibene:
> Let's merge then, I'll give it a look later

sounds good thank you!

Issue #831, opened 2023-05-22
nk.ppg_findpeaks  return a list of list with "bishop " method, but return a list with "elgendi" method 


Ian-Ren:
Run 

signals, info = nk.ppg_process(data, sampling_rate=25, method_cleaning='elgendi', method_peaks='elgendi')
nk.ppg_plot(signals, sampling_rate=25)

Everything is good
But run 
 signals, info = nk.ppg_process(data, sampling_rate=25, method_cleaning='elgendi', method_peaks='bishop')
nk.ppg_plot(signals, sampling_rate=25)

Errors is thrown. 

Further debug into the problem, found that the problem is that when call "ppg_findpeaks" function with 'elgendi' method,  the peaks are returned as a list. While calling with 'bishop' method, peaks are returned as a list of list.   Please squeeze the other dimension. 

Added  possible solutions: 
NeuroKit/neurokit2/ppg/ppg_findpeaks.py
Line 223   peaks = np.asarray(np.where(m_max_sum == 0)).astype(int)    -->  peaks =np.where(m_max_sum == 0)[0].astype(int)
Line 224    onsets = np.asarray(np.where(m_min_sum == 0)).astype(int)  --> onsets = np.where(m_min_sum == 0)[0].astype(int)

I am not proficient python programmer. You can do better to fix this

First time to post issue on Github.  Hope I stated it clearly enough. Let me know if there is anything unclear. 

Thanks 

Ian

PR #833, opened 2023-05-30
[Fix] bishop peaks format

danibene:
# Description

This PR aims at fixing an issue with the output format of PPG peaks returned with the bishop method. See also https://github.com/neuropsychology/NeuroKit/issues/831

# Proposed Changes

As suggested by @Ian-Ren (thank you!) I changed the `ppg_findpeaks()` function so that only one dimension is returned. 
I also modified the `test_ppg_report()` function so that the bishop method is tested, to confirm that this change does make the output format correct.

# Checklist


DominiqueMakowski:
looks good to me

PR #841, opened 2023-06-14
[Feature] EMG report

danibene:
# Description

This PR aims to generate an interactive HTML report describing EMG processing as described in https://github.com/neuropsychology/NeuroKit/issues/665

# Proposed Changes

1. I changed the `emg_plot()` function so that it optionally generates an interactive plot with plotly
2. I added an `emg_methods()` function
3. I changed the `emg_process()` function so that it optionally generates a report
5. I modified the docstring in `report.py` to reference EMG; however, I did not add any EMG-specific logic for generating a summary table. I wonder if we should use `bio_analyze()` to generate the table rather than checking the column names within `report.py`

# Checklist

### DIFF
--- a/neurokit2/emg/emg_clean.py
+++ b/neurokit2/emg/emg_clean.py
@@ -9,7 +9,7 @@ from ..misc import NeuroKitWarning, as_vector
 from ..signal import signal_detrend
 
 
-def emg_clean(emg_signal, sampling_rate=1000):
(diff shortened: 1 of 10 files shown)


danibene:
The test is currently failing because I returned a figure even when it is generated with matplotlib to keep the option of generating a report with matplotlib instead of plotly

DominiqueMakowski:
can we merge that?

Issue #853, opened 2023-06-28
Add Respiration Amplitude Variability (RAV) ?


DominiqueMakowski:
I am not aware of any literature about that (didn't even check), but intuitively it would make sense.

We could add it as part of `rsp_rrv()`, adding an `amplitude=None` argument, and if one passes the amplitude signal (we would do that in rsp_analyze), it computes the RAV_RMSSD (we currently compute the RSP_Amplitude_Mean in `intervalrelated()`)

PR #873, opened 2023-08-01
[Feature] Respiratory Amplitude Variability (RAV)

DominiqueMakowski:
#853 

I'm not aware of much literature on that, but it doesn't seem absurd as a potential feature of interest


danibene:
@DominiqueMakowski just had a quick look so far - do you think you could add some tests?

PR #882, opened 2023-08-11
[Feature] download zip

danibene:
# Description

This PR aims to allow for zip files to be automatically downloaded and extracted, as a first step towards automating the downloading and formatting of various datasets (see https://github.com/Tam-Pham/HRVStructure/issues/2#issuecomment-1656098785)

# Proposed Changes

I added a `download_zip()` function to a `database.py` file (was thinking that this could be where the refactored functions for downloading individual databases could be moved to within NeuroKit), and I used this function for the fantasia and mit arrhythmia databases.

Let me know if you want me to change the naming (wasn't sure about `database.py`), and also if you have any thoughts on whether there should be automated tests for this (not sure if downloading entire datasets is a good idea with GitHub Actions)

# Checklist

### DIFF
--- a/data/fantasia/download_fantasia.py
+++ b/data/fantasia/download_fantasia.py
@@ -13,7 +13,7 @@ import pandas as pd
 import numpy as np
 import wfdb
 import os
-from pathlib import Path
+import pathlib
 
 import neurokit2 as nk
 
@@ -39,7 +39,7 @@ dfs_rpeaks = []
 
 for i, participant in enumerate(files):
 
-    data, info = wfdb.rdsamp(str(Path(database_path, participant)))
+    data, info = wfdb.rdsamp(str(pathlib.Path(database_path, participant)))
 
     # Get signal
     data = pd.DataFrame(data, columns=info["sig_name"])
@@ -50,7 +50,7 @@ for i, participant in enumerate(files):
     data["Database"] = "Fantasia"
 
     # Get annotations
-    anno = wfdb.rdann(str(Path(database_path, participant)), 'ecg')
+    anno = wfdb.rdann(str(pathlib.Path(database_path, participant)), 'ecg')
     anno = anno.sample[np.where(np.array(anno.symbol) == "N")[0]]
     anno = pd.DataFrame({"Rpeaks": anno})
     anno["Participant"] = "Fantasia_" + participant
--- a/data/fantasia/download_fantasia.py
+++ b/data/fantasia/download_fantasia.py
@@ -15,14 +15,14 @@ import wfdb
 import os
 from pathlib import Path
 
-from neurokit2.data import download_zip
+import neurokit2 as nk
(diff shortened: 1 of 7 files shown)


data/fantasia/download_fantasia.py
```diff
@@ -13,9 +13,24 @@
 import numpy as np
 import wfdb
 import os
+from pathlib import Path
 
+from neurokit2.data import download_zip
```
DominiqueMakowski, line 18: can we use the standard `import neurokit2 as nk` and then `nk.download_zip()`?

neurokit2/data/database.py
```diff
@@ -0,0 +1,53 @@
+import requests
+import zipfile
+from pathlib import Path
```
DominiqueMakowski, line 3: Same here can we do `import pathlib` and then `pathlib.Path` (I like to see where functions come from ^^)

danibene: I am the same way and did this in my previous code but then I saw pretty much everyone else using `from pathlib import Path` and so switched for the sake of convention, up to you though

DominiqueMakowski: yeah let's be iconoclast and internally consistent ^^

neurokit2/data/database.py
```diff
@@ -0,0 +1,53 @@
...
+
+def download_zip(url, destination_path):
```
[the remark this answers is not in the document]
danibene: ah I just realized you meant downloading any files from a url and not just a zip. Yeah sure I can do that too (though I feel like then we should still have `download_zip()` that calls `download_from_url()`, since there are zip-specific functionalities in the function like removing the original zip file), but I have to go now, can get back to it in a few days probably

neurokit2/data/database.py
```diff
@@ -0,0 +1,114 @@
+import requests
+import zipfile
+import pathlib
+import urllib.parse
+
+def download_from_url(url, destination_path=None):
+    """
+    Download a file from the given URL and save it to the destination path.
+
```
DominiqueMakowski, line 9:
```suggestion
    """**Download Files from URLs**
    
    Download a file from the given URL and save it to the destination path.

```

neurokit2/data/database.py
```diff
@@ -0,0 +1,115 @@
...
+        with destination_path.open("wb") as file:
+            file.write(response.content)
+        return True
+    else:
+        return False
+    
+
+def download_zip(url, destination_path=None, unzip=True):
+    """Download a ZIP file from a URL and extract it to a destination directory.
+
```
DominiqueMakowski, line 39:
```suggestion
    """**Download ZIP files**
    
    Download a ZIP file from a URL and extract it to a destination directory.
    
```

DominiqueMakowski:
Looks good to me, should we merge?

PR #883, opened 2023-08-13
[Feature] PPG plots improvements

DominiqueMakowski:
## Breaking
- `ecg_plot()` and `ppg_plot()` now *require* the sampling_rate to be passed (for segmentation to happen) - also simplifies the code. 

## Changes

- [x] Add `ppg_peaks()` master function to bring it up-to-speed with ECG
- [x] Add `ppg_segment()` to be able to segment individual peaks and nicely visualize them
- [x] Improve `ppg_plot()` by adding the individual peaks overlays to the plot (like ECG)

image

### DIFF
--- a/neurokit2/ecg/ecg_plot.py
+++ b/neurokit2/ecg/ecg_plot.py
@@ -8,7 +8,7 @@ from ..ecg import ecg_peaks
 from ..epochs import epochs_to_df
 from ..signal import signal_fixpeaks
 from ..stats import rescale
-from .ecg_segment import _ecg_segment_plot, ecg_segment
+from .ecg_segment import ecg_segment
 
 
 def ecg_plot(ecg_signals, rpeaks=None, sampling_rate=1000, show_type="default"):
--- a/neurokit2/ecg/ecg_plot.py
+++ b/neurokit2/ecg/ecg_plot.py
@@ -5,7 +5,6 @@ import numpy as np
 import pandas as pd
 
 from ..ecg import ecg_peaks
-from ..epochs import epochs_to_df
 from ..signal import signal_fixpeaks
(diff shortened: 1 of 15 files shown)


neurokit2/ecg/ecg_plot.py
```diff
@@ -138,40 +140,56 @@ def ecg_plot(ecg_signals, rpeaks=None, sampling_rate=None, show_type="default"):
...
         ax1.legend(loc="upper right")
 
         # Plot individual heart beats.
-        if sampling_rate is not None:
-            ax2.set_title("Individual Heart Beats")
+        # TODO: how can we directly insert the figure from ecg_segment?
+        # fig_beats = ecg_segment(
+        #     ecg_signals["ECG_Clean"], peaks, sampling_rate, show="return"
+        # )
+        # ax2 = fig_beats.axes[0]
```
DominiqueMakowski, line 169: Now `ecg_segment(..., show="return")` returns a Figure, but how can we insert that figure into the axis so that we don't have to re-plot it? 🤔  @danibene any ideas from the top of your head?

<sub>*matplotlib is the worse*</sub>

danibene: Not off the top of my head, if I understood your question correctly it seems like it's not possible to insert axes from another figure? https://stackoverflow.com/questions/66369315/how-to-assign-axes-from-one-figure-to-axes-from-another-figure

DominiqueMakowski: ok thanks, that was helpful, when with the `ax` kwarg similar to other packages

DominiqueMakowski:
@danibene I'm not sure how / if we can transfer that to the dynamic plots, but I'd say let's merge for now and then we'll see for the rest

PR #889, opened 2023-08-22
[Breaking] for *_plot() and ecg_fixpeaks(method="neurokit")

DominiqueMakowski:
- [x] **Breaking**: Now `ecg_fixpeaks(method="neurokit")` returns also a tuple with an info dict for consistency with kubios method
- [x] **Breaking**: `ecg_plot()` loses the show_type="artifacts" option 
- [x] **Breaking**: `*_plot()` now take an "info" dict as input instead of the `sampling_rate` argument
- [x] ecg_plot improvements
  - [x] Move the ax0 plot code of `ecg_plot()` into a separate internal function in `ecg_peaks()` + potential improvement of showing the "fixed" peaks (would also help removing the weird `show_type` option that we have in `ecg_plot()`)
  - [x] ~Add delineation plot in ecg_plot, as part of `ecg_segment_plot` code?~
- [x] Add show methods (with `ax` arg) to ecg_rate, ecg_delineate, ecg_peaks that can then get called from `ecg_plot()` that would just to an assembling of existing plots, to remove code complexity and duplication
- [x] ecg_clean : method based on convolution of the simulated wavelet over the signal?

Roadmap

### RSP

### PPG
- [ ] Test the viability of running `fixpeaks` as part of `ppg_peaks()` routine

### Misc
- [ ] ecg_peaks() plot: Add non-corrected peaks (+ ectopic, extra, missing etc.?)


DominiqueMakowski:
New `ecg_plot()` with additional info

image

danibene:
> - [ ] Dynamic plots: @danibene instead of re-plotting with plotly, can't we just turn the existing matplotlib figs as interactive in htmls?

I did look into this at some point and I remember there was a library for this, I don't have time to find my old notebook right now, but I think it was https://mpld3.github.io/modules/API.html#mpld3.fig_to_html

Issue #593, opened 2021-12-10
ECG Analysis bug for sample rate of 500


sametdumankaya:
I wanted to analyze my ECG data (sample rate of 500), however I came across with this bug and I don't know how to update the parameter 'windows'.

Code:
`processed_data, info = nk.bio_process(ecg=df["V3"], sampling_rate=500)`
`results = nk.bio_analyze(processed_data, sampling_rate=500)`

Error:
`ValueError: NeuroKit error: the window cannot contain more data points than the time series. Decrease 'windows'.`

DominiqueMakowski:
Hi, could you attach a piece of reproducible code and data for us to see what's going on? Thanks

Soumadip-Saha:
Hello @DominiqueMakowski, I am also facing the same issue for the 500Hz sample frequency.
`def SWT_calculate(data, level):
    db1 = pywt.Wavelet('db3')
    x = data[1,:].tolist()
    x = np.pad(x, (2, 2), 'constant', constant_values=(0, 0))
    coeff = pywt.swt(x, db1, level=level)
    lead = np.square(coeff[0][1])
    lead = (lead - np.min(lead))/np.ptp(lead)
    lead = lead[lead!=0]
    ent = -np.dot(np.log(lead), lead)
    return ent

def Create_wide_features(data):
    df, info = nk.ecg_process(data[1], sampling_rate=500)
    R_peaks = info['ECG_R_Peaks']
    temp_df = nk.ecg_analyze(df,sampling_rate=500)`

ZacDair:
Having the same issue utilising the WESAD dataset PPG data (64Hz) windowed into 10 second segements.

anshu-97:
Hi @ZacDair , apologies for the late reply! Could you please attach your code as well as data so that we can try reproducing the error?

zara6697:
Hi,
I have the same issue. Is there any solution for it?
Best regards.

zara6697:
Hi everyone, @[ZacDair](https://github.com/ZacDair), @[Soumadip-Saha](https://github.com/Soumadip-Saha)
I found the source of this error. when the number of 'RRI' in different window is not the same, this error is shown. to solve it, you could replace the '_hrv_dfa' with the following code:
def _hrv_dfa(peaks, rri, out, n_windows="default", **kwargs):

    if "dfa_windows" in kwargs:
        dfa_windows = kwargs["dfa_windows"]
    else:
        print(rri.shape)
[...]

PR #901, opened 2023-09-12
0.2.7

DominiqueMakowski:
Roadmap

### RSP

### PPG

### Misc

### Complexity

- [x] Add quadratic sample entropy? 
  - https://umu.diva-portal.org/smash/get/diva2:1068270/FULLTEXT01.pdf 
  - https://github.com/kirajcg/pysampen/blob/master/pysampen/sampen.py#L79

### DIFF
--- a/docs/functions/complexity.rst
+++ b/docs/functions/complexity.rst
@@ -80,11 +80,11 @@ Fractal Dimension
 Entropy
 ^^^^^^^^^^^^^^^^^
 *entropy_shannon()*
-"""""""""""""""""""
+"""""""""""""""""""""
 .. autofunction:: neurokit2.complexity.entropy_shannon
 
 *entropy_maximum()*
-"""""""""""""""""""
+"""""""""""""""""""""
 .. autofunction:: neurokit2.complexity.entropy_maximum
 
 *entropy_differential()*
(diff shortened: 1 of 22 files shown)


DominiqueMakowski:
@danibene the docs CI flagged the `intervals_process()` function in hrv/intervals_process, but looking at it 1) I'm super suprised it hasn't been detected before because the example likely never ran; 2) the example seems weird; 3) it makes me wonder whether this should even be an exported function, or whether we should turn it into an internal `_preprocess_rpeaks_intervals()` 

any thoughts?

danibene:
> @danibene the docs CI flagged the `intervals_process()` function in hrv/intervals_process, but looking at it 1) I'm super suprised it hasn't been detected before because the example likely never ran; 2) the example seems weird; 3) it makes me wonder whether this should even be an exported function, or whether we should turn it into an internal `_preprocess_rpeaks_intervals()` 
> 
> any thoughts?

1) yeah I'm not sure either why it's being flagged now
2) what about the example seems weird? Maybe we could show how it can be used prior to the hrv functions? (So that the R-R intervals are given as an input?)
3) I don't think so, this would then prevent processing the R-R intervals prior to time-domain or nonlinear HRV feature computation. Though admittedly, this is probably not well known/documented enough for users to be using it this way

PR #917, opened 2023-10-02
[Fix] `SCR_Peaks_Amplitude_Mean` should be only computed on peaks

mperreir:
# Description

For the moment `eda_intervalrelated()` computes the mean peaks amplitude based on the whole signal (where most of the amplitude values are zero). This leads to very low (and wrong) `SCR_Peaks_Amplitude_Mean` values.

# Proposed Changes

Compute the mean peaks amplitude only on the peaks (and not on the whole signal). For that we first select rows where `SCR_Peaks` has a value of 1.

# Checklist


DominiqueMakowski:
The docstring example now fails I think

https://github.com/neuropsychology/NeuroKit/actions/runs/6382716523/job/17321902504?pr=917#step:5:99

mperreir:
Yes, it seems. I was not expecting that. It might be fixed with something like this:
```python
only_peaks = data["SCR_Peaks"] == 1
if len(only_peaks) > 0:
    output["SCR_Peaks_Amplitude_Mean"] = np.nanmean(data[data["SCR_Peaks"] == 1]["SCR_Amplitude"].values)
else:
    output["SCR_Peaks_Amplitude_Mean"] = 0
```
I'll try to push this tomorrow.

PR #922, opened 2023-10-28
Update ecg_segment.py

hyunlimy:
modify plotting error

# Description

I think there was an error referencing the `y` value incorrectly when drawing the segment_plot.
Since the previously declared `mean_heartbeat` is a grouped DataFrame, you need to explicitly specify the columns to get the values correctly.

# Proposed Changes

I changed the `_ecg_segment_plot()` function so that it can properly referencing the `y` value.


DominiqueMakowski:
Hi @hyunlimy, may I ask why did you close this?

PR #926, opened 2023-11-06
0.2.8

DominiqueMakowski:
Roadmap

### RSP

### PPG

### Misc


setup.py
```diff
@@ -30,7 +28,7 @@ def find_version():
 requirements = ["numpy", "pandas", "scipy", "scikit-learn>=1.0.0", "matplotlib"]
 
 # Optional Dependencies (only needed / downloaded for testing purposes, for instance to test against some other packages)
-setup_requirements = ["pytest-runner", "numpy"]
+setup_requirements = ["pytest-runner", "numpy", "requests"]
```
DominiqueMakowski, line 31: is that alright @hoechenberger ?

hoechenberger: Hello, no, we need it in `requirements` :) Is my current understanding

hoechenberger: Great, thanks!

Issue #938, opened 2023-12-12
Accelerated Visibility-Graph-Based Peak detector


taulokoka:
Dear NeuroKit,

the visibility graph method you have included in your package (koka2022) has been recently updated (https://ieeexplore.ieee.org/document/10290007) and is being maintained at https://github.com/JonasEmrich/vg-beat-detectors. It would be appreciated, if you could update the method since the recent version is orders of magnitude faster than the previous, while providing better results.

Best regards,
Taulant Koka

DominiqueMakowski:
Hi @JonasEmrich, there is a request for adding VG-based peak detection methods to NeuroKit, and I was wondering what your thoughts on this are ☺️ Options include: 

1. Doing nothing
    - Cons: No potential visibility gains (methods included in larger libraries like NK tend to be discovered more, and thus used and cited
2. Soft dependency: calling the vg-method from NK would call `vg-beat-detectors`, but since we operate on a minimum dependency, it would be a soft dependency, i.e., asking users to manually install the package the first time. 
    - Pros: no duplicated code, only one place to maintain
    - Cons: More prone to potential breaks (as changes your end can break things upstream). + many users cannot necessarily install smaller packages used for a couple of functions (e.g., in production contexts). That point likely doesn't apply as even with the other solutions below we would still need a conditional dependency on `ts2vg`
3. Duplicated code (code giveaway): You agree for your code to be copied over and eventually adapted
    - Cons: duplicated code. Increased maintenance burden for non-experts (on NK side)
4. Duplicated code (preserved maintainership): only difference is that you become a contributor of the adaptation/implementation of the method in NK and can help us out with maintenance 
    - Pros: you are still responsible for the code, you become part of the NK team. More stable (i.e., you can still make quick updates to your library without fears of breaking NK, which can use a more stable/slow paced updating process) 
    - Cons: duplicated code to maintain 
5. Full integration into NK
    - Cons: less visibility for your work as it becomes part of NK

I'd say that option 4 is one of the best in terms of pros/cons balance, but let us know what you think!

JonasEmrich:
Hi @DominiqueMakowski, thanks for reaching out! 
I also think option 4 would be the optimal choice. Since I am fairly new to the process, maybe you could guide a bit me through the next steps 😃

DominiqueMakowski:
Of course :)

The rough steps would be:

1. Fork NeuroKit
2. Put yourself on the `dev` branch
[...]

PR #939, opened 2023-12-12
[Feature] Add Fast Visibility Graph Based ECG R-peak Detector

JonasEmrich:
Adding the accelerated visbility-graph-based R-peak detector (https://ieeexplore.ieee.org/document/10290007, https://github.com/JonasEmrich/vg-beat-detectors) #938


neurokit2/ecg/ecg_findpeaks.py
```diff
@@ -105,12 +97,12 @@ def _ecg_findpeaks_findmethod(method):
         return _ecg_findpeaks_rodrigues
     elif method in ["vg", "vgraph", "koka2022"]:
         return _ecg_findpeaks_vgraph
+    elif method in ["FastNVG", "fastnvg", "emrich", "emrich2023"]:
```
DominiqueMakowski, line 100: `method = method.lower()` above so you don't need the upper case variant :)

JonasEmrich: True. I have fixed it in [7674b25](https://github.com/neuropsychology/NeuroKit/pull/939/commits/7674b25f5b3d5d44f4c6cad82c77f3531d9c9588)

JonasEmrich:
Now, I should have adapted the package at all the required places. Together with the authors of the koka2022 method, we decided to remove it in favor of this more advanced method. 

I tested the implementation briefly on two databases, and it appears that it procures results comparable to the original FastNVG algorithm (https://github.com/JonasEmrich/vg-beat-detectors)

DominiqueMakowski:
Amazing job! Thanks

> Together with the authors of the koka2022 method, we decided to remove it in favor of this more advanced method.

In that case, could we leave the `"koka2022"` input to then run your udpated method, throwing a warning that `"The 'koka2022' method has been replaced by XX. Please replace method='koka2022' by ..."` (so that we don't break potential user code just yet)

JonasEmrich:
> In that case, could we leave the `"koka2022"` input to then run your udpated method, throwing a warning that `"The 'koka2022' method has been replaced by XX. Please replace method='koka2022' by ..."` (so that we don't break potential user code just yet)

Thanks for pointing this out! I added a warning in `ecg_clean()` and `ecg_findpeaks()`.

DominiqueMakowski:
I think we can just add it in `ecg_findpeaks()` because it's there that the switch happens

JonasEmrich:
> I think we can just add it in `ecg_findpeaks()` because it's there that the switch happens

Although the emrich2023 method uses the same cleaning as koka2022 (points to the same function), I think that perhaps users also should use emrich2023 here instead of koka2022. It's just the naming but makes it more consistent, since the koka2022 method is deprecated/removed. Or what are your thoughts?

DominiqueMakowski:
> It's just the naming but makes it more consistent, since the koka2022 method is deprecated/removed

makes sense!

DominiqueMakowski:
@JonasEmrich is it good for me to merge?

JonasEmrich:
> @JonasEmrich is it good for me to merge?

@DominiqueMakowski yes, from my side it is ready to merge

DominiqueMakowski:
Amazing, thanks again for all the work :)

NikkiNgombe:
Hello @JonasEmrich and @DominiqueMakowski ,
I'm working with some ECG and wanted to use "emrich2023" for the peak detection. However, I get the following error:

ValueError: NeuroKit error: ecg_findpeaks(): 'emrich2023' not implemented.

After installing ts2vg, I was able to use koka2022, but I am still unable to run ecg_findpeaks() using the Emrich 2023 method. Is this emrich2023 method actually implemented in NeuroKit2? When I checked the source code for ecg_findpeaks(), I also noticed, that emrich2023 wasn't listed in the available methods.
Screenshot 2024-07-02 at 20 30 42

Any help would be greatly appreciated, thank you!
Nikki

DominiqueMakowski:
I can see the lines in the current version: https://github.com/neuropsychology/NeuroKit/blob/1aa8deee392f8098df4fd77a23f696c2ff2d29db/neurokit2/ecg/ecg_clean.py#L126-L127

Do you have the latest version of NK installed?

NikkiNgombe:
Thank you for the quick response! I was using version 0.2.7. I have now updated to version 0.2.9 and can now use the emrich2023 option. Thank you so much! :)

Issue #823, opened 2023-05-12
ecg_peaks (method='neurokit') sometimes fails on ECG's with missing data


tecamenz:
**Describe the bug**
On certain ECG's with missing data or longer periods of constant values (e.g. zeros), `ecg_peaks` with `method="neurokit"` fails to find all but two peaks. Below I attached a screenshot with `show=True`:
![nk_fail](https://github.com/neuropsychology/NeuroKit/assets/52130886/a62f41bd-0cc2-4aad-af88-a8396622885d)

The cause of this issue lies in the way Neurokit calculates the `min_len`. This variable defines the minimum qrs length (`len_qrs`) required to accept a valid R-peak. 
https://github.com/neuropsychology/NeuroKit/blob/3d5ecfca8e9c10a1f4150f8877aba8f34c0d0a22/neurokit2/ecg/ecg_findpeaks.py#L249-L252

Changing `np.mean()` to `np.median` would provide a more robust estimate for the most common qrs length and would prevent `min_len` from becoming too high. The following screenshot shows the detected R-peaks with `np.median()`. 
Because the median calculation is a bit slower, I saw a 16% increase in runtime on a 24h ECG. 

![nk_pass_median](https://github.com/neuropsychology/NeuroKit/assets/52130886/e59b0479-4f94-4e01-831a-b49f8a1fbbba)

**System Specifications**
- OS: Linux ( 64bit) 
- Python: 3.8.10 
- NeuroKit2: 0.2.0 

- NumPy: 1.23.4 
- Pandas: 1.5.3 
- SciPy: 1.10.1 
- sklearn: 1.2.2 
- matplotlib: 3.6.2

DominiqueMakowski:
Interesting. 

@JanCBrammer what do you think about changing allowing for median instead of mean? 

I reckon we should probably do some benchmarking and some tests to see whether the median option is as good as mean in most cases

JanCBrammer:
@DominiqueMakowski, exactly my thoughts.
It's an interesting and useful observation. The median looks promising and on paper it's preferable over the mean in this context. 
But it'd be good to confirm that (i.e., the medians "superiority") empirically.

JanCBrammer:
Do we have benchmarking set up?
[...]

PR #940, opened 2023-12-14
[Feature] Add signal_flatintervals function for flatline detection

jchromik:
# Description

This PR aims at adding the function `signal_flatintervals`. The function can identify flatline intervals in signals and if modelled after `signal_flatline`. We discussed this feature in #823. It is one step towards making NeuroKit able to process long-term ECGs with bad or unprocessable intervals.

# Possible Next Step

A possible next step could be to a `bad_intervals` keyword argument to the peak detection functions to that the peak detector does not need to process the whole signal but only the good (non-flatline parts).

# Limitations / Concerns

- The function `signal_flatintervals` needs two "magic" parameters threshold and tolerance. `threshold` is the same as in signal_flatline with a default values of 0.01 (1% percent of the largest spike in the signal). `tolerance` says how long the flatline interval needs to be to be recognised as such default is 60 seconds. 
- The test uses `ecg_simulate` and is therefore quite slow (approx. 10s). I am open to suggestions on how to speed this up.

# Checklist

- [x] I ran the [CODE CHECKS](https://github.com/neuropsychology/NeuroKit/blob/master/.github/CONTRIBUTING.rst#run-code-checks) on the files I added or modified and fixed the errors. --> I did not correct complains in code that I did not touch
- [x] I have added the newly added features to **News.rst** (if applicable) --> not applicable


neurokit2/signal/signal_flatintervals.py
```diff
@@ -0,0 +1,118 @@
...
+    ----------
+    signal : Union[list, np.array]
+        The signal as a vector of values.
+    sampling_rate : int
+        The sampling rate of the signal, i.e. how many samples (values) per second.
+    threshold : float, optional
+        Flatline threshold relative to the biggest change in the signal.
+        This is the percentage of the maximum value of absolute consecutive differences.
+        Default: 0.01 (= 1% of the biggest change in the signal)
+    tolerance : int, optional
```
DominiqueMakowski, line 18: I would rename this for e.g., `duration_min`

jchromik:
@DominiqueMakowski 
I renamed the parameter `tolerance` to `duration_min`. All checks pass now. Code cov is complete for all my changes. 
(I'm still open for feedback and critique, though ;-) )

PR #943, opened 2024-01-04
[Improvement] Faster _ecg_findpeaks_elgendi

purpl3F0x:
# Description

Optimize _ecg_findpeaks_elgendi in those areas of the code:
1. `blocks` could be just a bool array
2. Detection of start/end of block by only comparing `prev` vs `cur` (`blocks[i-1]`, `blocks[i]`) value. (about 2x faster) (even better when combined with *"1"*)
3. Use `enumerate(zip(...))` for iterating the array, (gives a small performance -but measurable- improvement)
4. Move some thresholds outisde of the loop, not to get calculated at each iteration

# Checklist

### DIFF
--- a/neurokit2/ecg/ecg_findpeaks.py
+++ b/neurokit2/ecg/ecg_findpeaks.py
@@ -738,24 +738,24 @@ def _ecg_findpeaks_elgendi(signal, sampling_rate=1000, **kwargs):
     window2 = int(0.6 * sampling_rate)
     mwa_beat = _ecg_findpeaks_MWA(abs(signal), window2)
 
-    blocks = np.zeros(len(signal))
-    block_height = np.max(signal)
+    blocks = mwa_qrs > mwa_beat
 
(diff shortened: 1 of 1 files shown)


DominiqueMakowski:
For some reason it fails the tests

purpl3F0x:
It seems to be passing this time., worried-I did some copy-pasta typo on the variables. but seems ok now 🤷

DominiqueMakowski:
Out of curiosity, do you have a benchmark do show how much time we shave wit the version?

purpl3F0x:
> Out of curiosity, do you have a benchmark do show how much time we shave wit the version?

I did some measurments, mainly on individual parts of the code. Don't have the excact numbers in hand, but:
- the comparison optimization is 2x faster on average&worst case since it gets narrowed down from up to 4 to up to 2.
- On using the bool array, without taking into account the reduced memory footprint + not traversing the loop for finding it's max value (why was this needed??)
image

DominiqueMakowski:
Nice job, thanks!

PR #947, opened 2024-01-10
[Fix + Improvement] Faster _ecg_findpeaks_hamilton and bug-fix

purpl3F0x:
# Description

This PR aims to improve performanc of  `_ecg_findpeaks_hamilton` and fix some probable bugs

# Proposed Changes

### Bug Fixes:
1. Changed possible bug popping from `n_pks` if `len(n_pks) > 8`, it should probably be `len(n_pks)`
2. Removed dead code of `i += 1` at the end of the loop

### Improvements:
1. changed loop iteration range to avoid checks inside the loop
2. use `insort` instead of appending& sorting in an already sorted array (tested about x300-500 times faster on arrays of 50 elements)
3. Use deques instead of manually pushing and popping to reduce constantly allocating new memory.

# Checklist

### DIFF
--- a/neurokit2/ecg/ecg_findpeaks.py
+++ b/neurokit2/ecg/ecg_findpeaks.py
@@ -4,6 +4,8 @@ import pandas as pd
 import scipy.signal
 import scipy.stats
 from warnings import warn
+from bisect import insort
+from collections import deque
 
 from ..signal import signal_findpeaks, signal_plot, signal_sanitize, signal_smooth, signal_zerocrossings
 from ..misc import NeuroKitWarning
@@ -331,12 +333,12 @@ def _ecg_findpeaks_hamilton(signal, sampling_rate=1000, **kwargs):
 
     ma[0 : len(b) * 2] = 0
 
-    n_pks = []
+    n_pks = deque([], maxlen=8)
     n_pks_ave = 0.0
-    s_pks = []
+    s_pks = deque([], maxlen=8)
     s_pks_ave = 0.0
     QRS = [0]
-    RR = []
+    RR = deque([], maxlen=8)
     RR_ave = 0.0
 
     th = 0.0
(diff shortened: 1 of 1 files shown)


purpl3F0x:
Not sure about bug fix 1, but the current code doesn't seem logical.

danibene:
Hi @purpl3F0x , thank you for this! Could you explain what kind of "buggy" behavior you saw that motivated fix 1? It would be ideal if you could add a test https://github.com/neuropsychology/NeuroKit/tree/master/tests demonstrating that your fix works

purpl3F0x:
> Collaborator

Hello, the controversial code is pretty small. On this line https://github.com/neuropsychology/NeuroKit/blob/3d004b4777c3e59b288f46b9f0f812ac1e04b5d4/neurokit2/ecg/ecg_findpeaks.py#L372

There is a check for the length of the `n_pks`, but this part of the code is doing append/pop&on `s_pks`. Those 2 buffers behave similarly, and they hold the last 8 recent peaks. They don't have a relation to each other. It is a typo, from the original repo the code was based from. There is a good possibility it works by accident cause at some point both array will top up and have legth of 8 elements.

danibene:
I see, would you mind opening an issue on the original repository then too? I see that someone already opened an issue about the code at the end of the loop you removed https://github.com/berndporr/py-ecg-detectors/issues/23 

Until it's fixed, I think you could just add comments stating that the implementation differs slightly from the original code

purpl3F0x:
> I see, would you mind opening an issue on the original repository then too? I see that someone already opened an issue about the code at the end of the loop you removed [berndporr/py-ecg-detectors#23](https://github.com/berndporr/py-ecg-detectors/issues/23)
> 
> Until it's fixed, I think you could just add comments stating that the implementation differs slightly from the original code

I will also open an issue on the original repo. (https://github.com/berndporr/py-ecg-detectors/issues/24)

If I'm not mistaken the way this "works", now is with `s_pks` having a max length depending on the input signal, until the length of `n_pks` becomes 8. I think that's what called an undefined behavior 😁.

purpl3F0x:
The issue has been confirmed and fixed on the original repo as of https://github.com/berndporr/py-ecg-detectors/commit/f5d949d505eab76e80b25faf976673a63925166f

berndporr:
Thanks guys. I've applied the fix also to my repo. I had to apply it manually from the diff as you have linted your version but should be in syn now again.

danibene (approved):
This looks good to me, thanks @purpl3F0x & @berndporr !

PR #960, opened 2024-02-05
[FIX] rr xcorr arange steps don't match

birm:
Line 128 uses step size of 0.25/50, but line 137 uses 0.25/60

I think this is a typo, but please close if I'm wrong!


DominiqueMakowski:
good catch, thanks @birm! and sorry for the delay in reviewing this

Issue #964, opened 2024-02-17
The sparsEDA API is not compatible with NK2


stasinos:
The _eda_phasic_sparsEDA() function (eda_phasic.py) returns a (driver, SCL, MSE) triple where the calling function eda_phasic() expects a (tonic, phasic) pair.

Since sparsEDA is an algorithm for estimating the tonic component, I think this function should return `tonic` and `eda_signal` - `tonic`

stasinos:
PR #966 has been merged, this can close

PR #965, opened 2024-02-17
Make sparsEDA return a tonic,phasic pair

stasinos:
The sparseEDA algorithm decomposes the EDA into three components: the `driver`, the `SCL` (tonic), and `MSE` (residual).
This is incompable with eda_phasic() which expects a (tonic,phasic) pair, see also Issue #964 .

This PR resolves Issue #964 by making _eda_phasic_sparsEDA() return a (tonic,phasic) pair where `tonic` is `SCL` and `phasic` is the difference between the original signal and the tonic component detected by sparsEDA,


stasinos:
Oops, sorry, targeted master instead of dev

PR #982, opened 2024-04-06
0.2.9

DominiqueMakowski:
Roadmap

### RSP

### PPG
  - [PyPPG](https://github.com/godamartonaron/GODA_pyPPG)

### Misc
- [ ] Readme looks ugly but this is a global issue (https://github.com/orgs/community/discussions/113792)


DominiqueMakowski:
I think we can merge it, since the last update didn't update on Pypi anyway (#984)

Issue #580, opened 2021-11-16
'No R-peak detected' is not handled in ecg_quality


CeliaBenquet:
### Description of the bug
Using ecg.ecg_quality, the function `ecg.ecg_findpeaks` is used. It works well for all signals but for a really noisy signal (hand-crafted signal using ecg.ecg_simulate, with artificial powerline noise on top of it - see image and data provided), it turns out that no peaks are detected, which leads to an empty list that is not handled properly in the rest of the code leading to all sorts of errors (`IndexError`, `ValueError`, `ZeroDivisionError`, ... errors obtained while modifying the code to handling the error). 

**Error obtained on original code:** 
```
NeuroKit/neurokit2/ecg/ecg_findpeaks.py", line 260, in _ecg_findpeaks_neurokit
    end_qrs = end_qrs[end_qrs > beg_qrs[0]]
IndexError: index 0 is out of bounds for axis 0 with size 0
```

 ### To Reproduce

1. Get the provided data (`data.csv`) in Material section: clean data obtained using `nk.ecg_simulate()` with heart rate of 110Hz and PowerlineNoise added on top of it (sinusoid of amplitude 50Hz, energy ratio of 977% of the initial clean signal)
2. Use `nk.ecg_quality()` with `sampling_rate=256`
3. See error

### Expected behaviour
At least: I would expect the empty list to be handled so that the reported heart rate is 0 and quality is set to bad. 
At best: However, seeing the shape of the signal, I would expect the function `ecg_findpeaks` to spot the peaks but the quality algorithm to still be able to determine that the signal is noisy, based on its other SQIs.

### Screenshots and material on the signal used for the bug
_Shape of the signal on which the bug occurs_
![Screen Shot 2021-11-16 at 12 06 53](https://user-images.githubusercontent.com/32598028/141975237-26b980cf-f96b-4f1d-a6f3-d882d7f57601.png)

_Data to us to reproduce the bug (data used for plot above)_
[data.csv](https://github.com/neuropsychology/NeuroKit/files/7545861/data.csv)

### System Specifications

- OS: Darwin ( 64bit) 
- Python: 3.9.7 
- NeuroKit2: 0.1.4.1 

- NumPy: 1.21.2 
- Pandas: 1.3.1 
- SciPy: 1.6.2 
- sklearn: 1.0.1 
- matplotlib: 3.4.2

zen-juen:
Hi @CeliaBenquet 

[...]

Issue #472, opened 2021-06-26
Meet an IndexError when using method ecg_peaks


skywalkerzhang:
I encounter the IndexError when use this(shown below) to get the QRS interval
```python
_, r_peaks = nk.ecg_peaks(ecg_signal, sampling_rate=200)
```
The error stack is shown as below:
```
Traceback (most recent call last):
  File "C:\Users\16778\anaconda3\lib\site-packages\IPython\core\interactiveshell.py", line 3437, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-2-5f81baaf3b97>", line 1, in <module>
    runfile('C:/Users/16778/Documents/ECG/python-classifier-2021/QRS_interval.py', wdir='C:/Users/16778/Documents/ECG/python-classifier-2021')
  File "C:\Users\16778\scoop\apps\pycharm-professional\current\IDE\plugins\python\helpers\pydev\_pydev_bundle\pydev_umd.py", line 197, in runfile
    pydev_imports.execfile(filename, global_vars, local_vars)  # execute the script
  File "C:\Users\16778\scoop\apps\pycharm-professional\current\IDE\plugins\python\helpers\pydev\_pydev_imps\_pydev_execfile.py", line 18, in execfile
    exec(compile(contents+"\n", file, 'exec'), glob, loc)
  File "C:/Users/16778/Documents/ECG/python-classifier-2021/QRS_interval.py", line 24, in <module>
    _, r_peaks = nk.ecg_peaks(ecg_signal, sampling_rate=200)
  File "C:\Users\16778\anaconda3\lib\site-packages\neurokit2\ecg\ecg_peaks.py", line 71, in ecg_peaks
    rpeaks = ecg_findpeaks(ecg_cleaned, sampling_rate=sampling_rate, method=method)
  File "C:\Users\16778\anaconda3\lib\site-packages\neurokit2\ecg\ecg_findpeaks.py", line 135, in ecg_findpeaks
    rpeaks = _ecg_findpeaks_neurokit(ecg_cleaned, sampling_rate, show=show)
  File "C:\Users\16778\anaconda3\lib\site-packages\neurokit2\ecg\ecg_findpeaks.py", line 260, in _ecg_findpeaks_neurokit
    end_qrs = end_qrs[end_qrs > beg_qrs[0]]
IndexError: index 0 is out of bounds for axis 0 with size 0
```

I guess this happened because my sample rate is 200.

[...]

Issue #134, opened 2020-02-28
ecg_process fails when sampling_rate low (e.g. 200Hz)


DominiqueMakowski:
```python
ecg90 = nk.ecg_simulate(duration=30, sampling_rate=200, heart_rate=90, random_state=42)
nk.ecg_process(ecg90)
```
```
nk.ecg_process(ecg90)
Traceback (most recent call last):

  File "<ipython-input-14-210b69211481>", line 1, in <module>
    nk.ecg_process(ecg90)

  File "D:\Dropbox\RECHERCHE\N\NeuroKit\neurokit2\ecg\ecg_process.py", line 53, in ecg_process
    instant_peaks, rpeaks, = ecg_peaks(ecg_cleaned=ecg_cleaned,

  File "D:\Dropbox\RECHERCHE\N\NeuroKit\neurokit2\ecg\ecg_peaks.py", line 64, in ecg_peaks
    rpeaks = ecg_findpeaks(ecg_cleaned, sampling_rate=sampling_rate,

  File "D:\Dropbox\RECHERCHE\N\NeuroKit\neurokit2\ecg\ecg_findpeaks.py", line 97, in ecg_findpeaks
    rpeaks = _ecg_findpeaks_neurokit(ecg_cleaned, sampling_rate,

  File "D:\Dropbox\RECHERCHE\N\NeuroKit\neurokit2\ecg\ecg_findpeaks.py", line 165, in _ecg_findpeaks_neurokit
    end_qrs = end_qrs[end_qrs > beg_qrs[0]]

IndexError: index 0 is out of bounds for axis 0 with size 0
```

DominiqueMakowski:
Seems to be caused by the low sampling rate.

JohnDoenut:
@DominiqueMakowski, could you with `nk.ecg_process(ecg90, sampling_rate=200)`?

DominiqueMakowski:
Oops 😅

JohnDoenut:
Happens to the best 😄

PR #1007, opened 2024-07-10
[Fix] Return an empty array rather than throwing an exception if no QRS dat…

ajb5d:
# Description

This PR aims fixes an error where running ecg_peakdetect on a signal without any QRS complexes would throw an exception rather than returning an empty array. There are a couple of issues (mostly closed due to inactivity) that relate to this: #580, #472, #134

# Proposed Changes

I changed the `_ecg_findpeaks_neurokit()` function so that the length of `beg_qrs` is checked before accessing the first element. If the array is empty then return an empty array rather than triggering a ValueException.

# Checklist

### DIFF
--- a/neurokit2/ecg/ecg_findpeaks.py
+++ b/neurokit2/ecg/ecg_findpeaks.py
@@ -274,6 +274,9 @@ def _ecg_findpeaks_neurokit(
     qrs = smoothgrad > gradthreshold
     beg_qrs = np.where(np.logical_and(np.logical_not(qrs[0:-1]), qrs[1:]))[0]
     end_qrs = np.where(np.logical_and(qrs[0:-1], np.logical_not(qrs[1:])))[0]
+
+    if len(beg_qrs) == 0:
+        return []
     # Throw out QRS-ends that precede first QRS-start.
     end_qrs = end_qrs[end_qrs > beg_qrs[0]]
 
--- a/neurokit2/ecg/ecg_findpeaks.py
+++ b/neurokit2/ecg/ecg_findpeaks.py
@@ -276,7 +276,8 @@ def _ecg_findpeaks_neurokit(
     end_qrs = np.where(np.logical_and(qrs[0:-1], np.logical_not(qrs[1:])))[0]
 
     if len(beg_qrs) == 0:
-        return []
+        return np.array([])
+    
     # Throw out QRS-ends that precede first QRS-start.
     end_qrs = end_qrs[end_qrs > beg_qrs[0]]
 
@@ -503,6 +504,14 @@ def _ecg_findpeaks_zong(signal, sampling_rate=1000, cutoff=16, window=0.13, **kw
     ret = np.pad(clt, (window_size - 1, 0), "constant", constant_values=(0, 0))
     ret = np.convolve(ret, np.ones(window_size), "valid")
 
+    # Check that ret is at least as large as the window
+    if len(ret) < window_size:
+        warn(
+            f"The signal must be at least {window_size} samples long for peak detection with the Zong method. ",
+            category=NeuroKitWarning,
+        )
+        return np.array([])
+
     for i in range(1, window_size):
         ret[i - 1] = ret[i - 1] / i
     ret[window_size - 1 :] = ret[window_size - 1 :] / window_size
@@ -638,7 +647,8 @@ def _ecg_findpeaks_christov(signal, sampling_rate=1000, **kwargs):
                 if len(RR) > 5:
                     RR.pop(0)
                 Rm = int(np.mean(RR))
-
+    if len(QRS) == 0:
+        return np.array([])
     QRS.pop(0)
     QRS = np.array(QRS, dtype="int")
     return QRS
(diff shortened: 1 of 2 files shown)


neurokit2/ecg/ecg_findpeaks.py
```diff
@@ -274,6 +274,9 @@ def _ecg_findpeaks_neurokit(
     qrs = smoothgrad > gradthreshold
     beg_qrs = np.where(np.logical_and(np.logical_not(qrs[0:-1]), qrs[1:]))[0]
     end_qrs = np.where(np.logical_and(qrs[0:-1], np.logical_not(qrs[1:])))[0]
+
+    if len(beg_qrs) == 0:
+        return []
```
danibene, line 279:
```suggestion
        return np.array([])
```
I would be inclined to return an empty numpy array so that the return type remains the same

danibene (approved):
Hi @ajb5d , thank you for this contribution! I would suggest also adding a [test](https://github.com/neuropsychology/NeuroKit/tree/master/tests) of the case when the array should be empty if possible. I'm happy to help you do this if you need any support.

DerAndereJohannes:
I see that you only put it on the neurokit method of the function i.e., `_ecg_findpeaks_neurokit`. Have you tested the other methods to see if they have the same result as your new version for neurokit? It would be nice to have consistent behaviour.

ajb5d:
@DerAndereJohannes @danibene I added tests for the other methods (at least checking that they don't crash when run on empty input). I made a couple of other fixes to ensure that most methods handle empty input. Two small things not addressed (But I wanted to record for posterity):
- I had initially added a check that they return no peaks when presented a flat array but a couple of the methods _do_ find peaks in an empty signal. That feels out of scope. 
- The `promac` method calls other methods and returns errors when no peaks are found by those methods. I think the root cause is `_ecg_findpeaks_promac_addconvolve` line 223: `mask[peaks] = 1` which fails when peaks is empty. I wasn't certain of the best way to handle this nested error so I'm leaving it untouched.

neurokit2/ecg/ecg_findpeaks.py
ajb5d: Fixed in the new commit

ajb5d:
I think the failed test is related to visbilitygraph which has a numpy dependency issue (we don't require a specific numpy version, the scipy > 1.13 is pulling in numpy 2.0 which isn't compatible with ts2vg). I couldn't get it to run locally on my machine either (but figured it was worth a shot to see if it ran in CI).

tests/tests_ecg_findpeaks.py
```diff
@@ -23,6 +24,23 @@ def _read_csv_column(csv_name, column):
...
+    failed_methods = []
+    for method in METHODS:
+        try:
+            method_func = _ecg_findpeaks_findmethod(method)
+            _ = method_func(np.zeros(12*240), sampling_rate=240)
+        except Exception:
+            failed_methods.append(method)
+            continue
+    if failed_methods:
+        raise Exception(f"Failed methods: {failed_methods}")
```
danibene, line 42: This also works but [parametrizing the test function](https://docs.pytest.org/en/stable/how-to/parametrize.html) could give a more detailed error message e.g.:
https://github.com/neuropsychology/NeuroKit/blob/781ebd6fe5d6546f2231fe0a3b8cdc78b3c4ac8c/tests/tests_rsp.py#L381-L389

tests/tests_ecg_findpeaks.py
```diff
@@ -23,6 +24,23 @@ def _read_csv_column(csv_name, column):
     csv_data = pd.read_csv(csv_path, header=None)
     return csv_data[column].to_numpy()
 
+def test_ecg_findpeaks_all_methods_handle_empty_input():
+    METHODS = ["neurokit", "pantompkins", "nabian", "gamboa", 
+               "slopesumfunction", "wqrs", "hamilton", "christov",
+               "engzee", "manikandan", "elgendi", "kalidas", 
+               "martinez", "rodrigues"]
```
danibene, line 31: If you decide to leave out the `"vgraph"` method for this PR, I would suggest specifying in a comment why you are not including it, maybe with a link to the comment in the PR

danibene (changes requested):
Thanks for making these changes! If you are unable to get the test working for the vgraph method, I think it is fine to merge without for now, as long as it is documented as a known issue.

ajb5d:
I added a comment and refactored the test method to use the parameterized methods.

tests/tests_ecg_findpeaks.py
ajb5d: fixed in last commit

tests/tests_ecg_findpeaks.py
ajb5d: comment added and opened #1009 to track the ts2vg issue.

tests/tests_ecg_findpeaks.py
```diff
@@ -23,6 +25,17 @@ def _read_csv_column(csv_name, column):
     csv_data = pd.read_csv(csv_path, header=None)
     return csv_data[column].to_numpy()
 
+#vgraph is not included because it currently causes CI to fail (issue 1007)
```
danibene, line 28:
```suggestion
#vgraph is not included because it currently causes CI to fail (issue #1009)    
```

danibene (approved):
Looks good to me!

DominiqueMakowski:
amazing, thanks all!

Issue #999, opened 2024-06-12
Images in documentation website not rendered


inezpereira:
Not a bug in the neurokit code per se, but on the documentation website.

🔸 **Images not being rendered**
Throughout the website, images are not being rendered. Example [here](https://neuropsychology.github.io/NeuroKit/functions/eda.html).

🔸 **To Reproduce**
Open [this page](https://neuropsychology.github.io/NeuroKit/functions/eda.html) and scroll down.

🔸 **Screenshots**

<img width="817" alt="image" src="https://github.com/neuropsychology/NeuroKit/assets/20027814/7a5c0005-17b0-49d2-91bc-fddf17c158c3">

🔸 **System Specifications**
I am on a MacOS Sonoma version 14.5. Documentation pages viewed on Brave and Safari browsers.

PR #1010, opened 2024-07-13
[Fix] Documentation Image Dependency

DerAndereJohannes:
The example images were not working for a while on the documentation website and I was curious as to what was wrong. It turns out that ipython at some point made the pickleshare dependency optional which meant that it was not automatically installed in the doc building workflow. I found the solution here on this issue ipython/ipython#14367. The solution is either to add pickleshare as a dependency or to update IPython. In this pull request I have simply gone with adding pickleshare as a dependency to get the images back online asap. Updating the remaining dependencies would be a topic for a different thread.

# Description

This PR fixes the issue #999 and allows users to be able to see the examples again on the documentation website. Here is the example from the issue with the actual image to see the result:
image-fix

# Proposed Changes

I added pickleshare as a dependency for the docs-build to allow for the doc image files to be stored in the correct location.

### DIFF
--- a/.github/workflows/docs-build.yml
+++ b/.github/workflows/docs-build.yml
@@ -44,6 +44,7 @@ jobs:
                   pip install EMD-signal
                   pip install cvxopt
                   pip install ts2vg
+                  pip install pickleshare
                   pip install https://github.com/neuropsychology/neurokit/zipball/dev
 
             - name: Build documentation 📜


DominiqueMakowski:
Hi @DerAndereJohannes sorry for the delay, thanks so much for that!!

PR #1014, opened 2024-07-21
[Fix] All docs-build warnings and errors

DerAndereJohannes:
# Description

In preparation for more documentation editing, I have dedicated this pull request to solving all the errors and warnings that still came up when performing the docs-build workflow. This also means that this pull request only consists of changes if they had a warning or not.

This means that many parts of the documentation now look more normal in design and that some ipython directives also are now able to run, creating more example images.

With these out of the way, it will now be easier to focus on new changes to the docs as we will immediately know if our new changes have caused warnings / errors.

# Proposed Changes

For every category of error / warning, I have created a commit including a comment as to what was done, why and what are the consequences of the change. The largest changes you can see in:
- signal_detrend
- progress_bar
- mne-to-df

I have also added the missing dependency that was previously updated in the docs-build workflow into the docs-check workflow.

The studies were giving the most warnings since most of the headers were duplicated in the files causing their reference to be turned into #id1, #id2 etc... rather than their section names. I firstly tried adding a specifier to leave the headings unchanged, only to change their id reference, however this was somehow not working. I think it could have to do that there is a problem when putting the markdown into the rst files.

The mne-to-df change is the only file that wasn't affected by some error / warning, but I changed the example to give a brief summary of the data since the function example was taking up the high majority of the eeg page.

Thanks.

# Checklist


DominiqueMakowski:
Thanks for all this work it's hugely appreciated!! Let me know when you want me to merge

(Please consider adding yourself as a [contributor](https://github.com/neuropsychology/NeuroKit/blob/master/AUTHORS.rst) for what it's worth)

DerAndereJohannes:
Happy to contribute. I have taken your advice and added another commit. 

Regarding this pull request, it can be pulled as is (which will fix the current issues in the docs). The pipeline now no longer gives errors or warnings, which you can see [here](https://github.com/neuropsychology/NeuroKit/actions/runs/10030258647/job/27719371739)

The next thing I want to do is 1. work on the codebook as discussed on #1012 and 2. fix more of the linking of the functions. I could also add this to this request (more relevant for number 2), but I could also create new pull requests at a later date.

Do let me know!

DominiqueMakowski:
Let's merge this bad boy now and make another PR for other things
Thanks again, and let me know if there's anything I can help with :)

PR #1021, opened 2024-08-16
[Fix] Prevent `sparseda` method from raising `AttributeError`

visserle:
>`np.Inf` was removed in the NumPy 2.0 release. Use `np.inf` instead.


DominiqueMakowski:
Nice catch, thanks!

Issue #1052, opened 2024-11-24
ecg_quality(method='averageQRS') is not bounded 


cgastaud:
**Describe the bug**
In `ecg_quality()` with method averageQRS, the result is expected to be an array between 0 and 1. However, since after rescaling the signal is interpolated with a quadratic method, this property no longer holds.
This can be easily solved by either choosing a linear interpolation method or clipping the result afterwards.

**To Reproduce**
`ecg = nk.ecg_simulate(duration=10, heart_rate=70, random_state=0)`
`print(nk.ecg_quality(ecg).min(), nk.ecg_quality(ecg).max())`

**Expected behaviour**
an array between 0 and 1

DominiqueMakowski:
That's a good point. I think we could use the same monotonic interpolation that we use for heart rate; or indeed, linear, which probably makes even more sense. Could you consider making a PR to implement the change?

PR #1044, opened 2024-10-25
[Change] Update ppg_quality.py to use consistent interpolation method ('previous')

athatcher13:
# Description

This PR aims at adding this feature...
Addressing issue #1043
Standardizing the interpolation step of the ppg_quality pipeline.

# Proposed Changes

Within ppg_quality, changed 'templatematch' implementation to use interpolation method 'previous' rather than 'quadratic'. This aligns it with the 'disimilarity' implementation and is a better representation of the beat-by-beat nature of the quality measure.

# Checklist

- [X] I have read the [CONTRIBUTING](https://github.com/neuropsychology/NeuroKit/blob/master/.github/CONTRIBUTING.rst#structure-and-code) file.
- [X] My PR is targeted at the **dev branch** (and not towards the master branch).
- [ ] I ran the [CODE CHECKS](https://github.com/neuropsychology/NeuroKit/blob/master/.github/CONTRIBUTING.rst#run-code-checks) on the files I added or modified and fixed the errors.
- [ ] I have added the newly added features to **News.rst** (if applicable)

PR #1054, opened 2024-11-30
[Change] ecg_quality interpolation from quadratic to "previous"

cgastaud:
# Description

Fix issue https://github.com/neuropsychology/NeuroKit/issues/1052: `ecg_quality` is unbounded because of quadratic interpolation.

# Proposed Changes

change interpolation to `linear`

# Checklist

### DIFF
--- a/neurokit2/ecg/ecg_quality.py
+++ b/neurokit2/ecg/ecg_quality.py
@@ -157,7 +157,7 @@ def _ecg_quality_averageQRS(ecg_cleaned, rpeaks=None, sampling_rate=1000):
 
     # Interpolate
     quality = signal_interpolate(
-        rpeaks, quality, x_new=np.arange(len(ecg_cleaned)), method="linear"
+        rpeaks, quality, x_new=np.arange(len(ecg_cleaned)), method="previous"
     )
 
     return quality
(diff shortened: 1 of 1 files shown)


neurokit2/ecg/ecg_quality.py
```diff
@@ -157,7 +157,7 @@ def _ecg_quality_averageQRS(ecg_cleaned, rpeaks=None, sampling_rate=1000):
 
     # Interpolate
     quality = signal_interpolate(
-        rpeaks, quality, x_new=np.arange(len(ecg_cleaned)), method="quadratic"
+        rpeaks, quality, x_new=np.arange(len(ecg_cleaned)), method="linear"
```
DominiqueMakowski, line 160:
```suggestion
        rpeaks, quality, x_new=np.arange(len(ecg_cleaned)), method="previous"
```

To make it consistent with https://github.com/neuropsychology/NeuroKit/pull/1044/files, maybe "previous" makes more sense

athatcher13: Agreed. While "previous" doesn't yield as nice-looking plots as other interpolation methods, I think it better represents the design intent. All PPG quality methods and the ECG quality method in question (averageQRS) measures are calculated per heartbeat, so it makes sense to have the quality value reported per heartbeat. The interpolation to match the length of the raw ECG/PPG signal is a convenience step.

PR #1072, opened 2025-02-11
Fix: nk.data import of eeg sample data

DerAndereJohannes:
# TL;DR
- Mne 1.9 changes how cropping works to include file checks (see https://github.com/mne-tools/mne-python/pull/12843 )
- Sample data is stored in pickle so does not come from disk
- Change removes previous existing filenames from eeg sample data

# Description

This PR fixes the documentation issue as one of the files uses the sample eeg data. The problem stems from the most recent version of mne which has a function which removes the filenames (and checks for their existence on disk!!) when trying to crop the data. The motivation here seems to be that it can remove the filenames if the crop range is out of scope of a file.

This caused issues, since the current version of the eeg sample data is imported via a pickle rather than eeg data itself. This included metadata that was relevant for the uploader, but of course the file name does not exist in this case.

Due to this, it was not possible to use the sample data and therefore the documentation pipeline could not work.

# Proposed Changes

The current proposal is to check for the version number and if it is greater than or equal to 1.9, then it removes the filenames from the eeg data. The change makes sense as there is in fact no filename connected to the in-memory file that came from the pickle, however, I am also not too sure if it looks too "hacky".

# Checklist

### DIFF
--- a/neurokit2/data/data.py
+++ b/neurokit2/data/data.py
@@ -4,7 +4,9 @@ import os
 import pickle
 import urllib
 
+import mne
 import pandas as pd
+
 from sklearn import datasets as sklearn_datasets
 
 
@@ -194,9 +196,7 @@ def data(dataset="bio_eventrelated_100hz"):
     # Dataframes ===============================
     if dataset == "iris":
         info = sklearn_datasets.load_iris()
-        data = pd.DataFrame(
-            info.data, columns=["Sepal.Length", "Sepal.Width", "Petal.Length", "Petal.Width"]
-        )
+        data = pd.DataFrame(info.data, columns=["Sepal.Length", "Sepal.Width", "Petal.Length", "Petal.Width"])
         data["Species"] = info.target_names[info.target]
         return data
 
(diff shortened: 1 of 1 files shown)


DerAndereJohannes:
It's the typical:

>It works on my machine!

I will take a closer look at it soon.

DominiqueMakowski:
I've updated the dataset on dev

```py
import pickle

import mne

raw = mne.io.read_raw_fif(
    mne.datasets.sample.data_path() / "MEG/sample/sample_audvis_raw.fif",
    preload=True,
    verbose=False,
)
raw = raw.pick(["eeg", "eog", "stim"], verbose=False)
raw = raw.crop(0, 60)
raw = raw.resample(200)

# raw.ch_names

# raw.info["sfreq"]

# Store data (serialize)
with open("eeg_1min_200hz.pickle", "wb") as handle:
    pickle.dump(raw, handle, protocol=pickle.HIGHEST_PROTOCOL)
```

and thought that would do the trick (rather than fixing in in the data function) but apparently not -_-

DerAndereJohannes:
The issue I think is that the `eeg_microstates` notebook still uses the .crop(0, 5) function on the data. Every crop on the data will check the existence of the file so the issue remains (check the filenames by doing `print(raw.filenames)` in your code).

It works on your machine, because the filenames are in fact correct to your setup. It's because of this I know that your username on your PC is "domma" ;-)

DominiqueMakowski:
oooooh right I see, good catch! I'll try to replace `.crop()` with our `mne_crop()` function which recreates a new object from scratch, maybe that'll help

Issue #1084, opened 2025-04-10
NeuroKit error: the window cannot contain more data points than the time series. Decrease 'scale'.


keysun11952:
**Describe the bug**
Calling ecg_analyze on PTB-XL dataset of 10-second 100Hz ECG result in this error:
NeuroKit error: the window cannot contain more data points than the time series. Decrease 'scale'.
This and related bugs has been raised multiple times in the past, all the way back to 2022, but never resolved.

**To Reproduce**
import numpy as np
import neurokit2 as nk
import matplotlib.pyplot as plt
folds = []
for f in range(10):
    data = np.load('data/fold_{}.npy'.format(f), allow_pickle=True)
    folds.append(data.item())
ecg = folds[0][21830][0]
signals, info = nk.ecg_process(ecg[:, 1], sampling_rate=100)
analysis = nk.ecg_analyze(signals, sampling_rate=100, method='interval-related')

DerAndereJohannes:
Hey keysun11952 thanks for submitting the issue. Sorry you have had this experience. The other issues you mention may have the same error, but for different contexts. I took a look at yours now.

In the source code it seems to fail for any signal where the r to r interval count is less than 12 will error out as it is not enough signal to compute the HRV dfa on the ecg signal.

The bug seems to happen in `_hrv_dfa` function where we compute the dfa windows to compute. 

```python3
    dfa_windows = kwargs.get("dfa_windows", [(4, 11), (12, None)])

    # Determine max beats
    if dfa_windows[1][1] is None:
        max_beats = (len(rri) + 1) / 10  # Number of peaks divided by 10
    else:
        max_beats = dfa_windows[1][1]
```
Since the default for the short window (the first tuple in the `dfa_windows` list goes from 4 to 11, it expects more 12 or more rri length. Some signals may not have 12 or more rri in a 10 second timeframe which is why you probably got the error on some signals and not on others.

I have proposed that we simply return nan if the rri length is less than 12. This would solve the issue and also as the hrv dfa is only really useful for very long signals, then we would never even really use this value anyway for shorter signals. 

See the pull request here: https://github.com/neuropsychology/NeuroKit/pull/1085


I would be interested to what you think.
Thanks.

DerAndereJohannes:
A fix has been implemented and merged into the `dev` branch. This means the changes will be included in the next official release.

If you would like to access the updates immediately, you can install the development version using the following command:

```sh
[...]

PR #1085, opened 2025-04-10
[Fix] Disable HRV DFA computation on signals with less than 12 rri

DerAndereJohannes:
# Description
This PR aims at fixing https://github.com/neuropsychology/NeuroKit/issues/1084 where if a signal has less than 12 rri, the short window for the HRV DFA computation fails. This pull suggests simply returning nan if we have less than 12 rri as the HRV DFA is not useful for short signals.

Please let me know what you think.


DominiqueMakowski:
Yes happy with that!!

PR #1104, opened 2025-06-18
Updated Elgendi reference in ppg_clean.py

peterhcharlton:
# Description

This PR aims to add a reference for the "elgendi" method used in "ppg_clean", as previously no reference was provided.

# Proposed Changes

I changed `ppg_clean.py` to add a reference to the following article:

_M. Elgendi, I. Norton, M. Brearley, D. Abbott, and D. Schuurmans (2013). Systolic peak detection in acceleration photoplethysmograms measured from emergency responders in tropical conditions. PLoS ONE, 8(10), 1–11._

I made changes to the functions `ppg_clean()` and `_ppg_clean_elgendi()`.

I presume that this is the correct [reference](https://doi.org/10.1371/journal.pone.0076585) because:
1. this is the reference used in `ppg_peaks.py`; and
2. it reports a very similar filtering approach (also using a Butterworth filter of bandpass of 0.5-8Hz, although neurokit currently uses order 3, whereas the original article reports order 2: "A zero-phase second-order Butterworth filter, with bandpass 0.5–8 Hz").

# Checklist

### DIFF
--- a/neurokit2/ppg/ppg_clean.py
+++ b/neurokit2/ppg/ppg_clean.py
@@ -63,6 +63,9 @@ def ppg_clean(ppg_signal, sampling_rate=1000, heart_rate=None, method="elgendi")
     * Nabian, M., Yin, Y., Wormwood, J., Quigley, K. S., Barrett, L. F., & Ostadabbas, S. (2018).
       An open-source feature extraction tool for the analysis of peripheral physiological data.
       IEEE Journal of Translational Engineering in Health and Medicine, 6, 1-11.
+    * M. Elgendi, I. Norton, M. Brearley, D. Abbott, and D. Schuurmans (2013).
+      Systolic peak detection in acceleration photoplethysmograms measured from emergency responders
+      in tropical conditions. PLoS ONE, 8(10), 1–11.
 
     """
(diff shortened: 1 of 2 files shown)


peterhcharlton:
If we agree that this is the correct reference, then we may want to update the code to use a Butterworth filter of order 2, rather than the current Butterworth filter of order 3.

peterhcharlton (commented):
This change aimed at resolving the failing test as a result of updating the ppg_clean filtering procedure.

I have increased the acceptable tolerance between the ground truth and estimated heart rates in `test_ppg_simulate`, from 15 to 20%. On doing so, all tests pass.

peterhcharlton:
Apologies, I've had a look at the failing tests, tried to resolve one (as stated above), and am at a bit of a loss as to how to resolve the remainder.

DominiqueMakowski:
Seems unrelated to that PR, I'll have a look

DominiqueMakowski:
(also @DerAndereJohannes in case you have some time to spare ☺️ )

DominiqueMakowski:
Same test fails on dev too, so I'll merge this anyway

PR #1150, opened 2025-11-25
 Fix pyxdf import error + enhance selective dejitter handling

Janet-king:
This PR fixes a flake8 error caused by pyxdf being imported inside a try block and later referenced.
It also improves the dejitter_timestamps parameter by allowing selective stream-level dejittering, avoiding double reads and unnecessary computation.

### DIFF
--- a/.github/workflows/docs-build.yml
+++ b/.github/workflows/docs-build.yml
@@ -59,5 +59,5 @@ jobs:
             - name: Deploy 🚀
               uses: peaceiris/actions-gh-pages@v3
               with:
-                  github_token: ${{ secrets.GITHUB_TOKEN }}
+                  github_token: ${{ secrets.GH_PAT }}
                   publish_dir: ./docs/_build


DerAndereJohannes:
Hi Janet-king, thank you very much for taking the time to create a contribution. 

## Notes on the contribution implementation
The gemini-code-assistant has given some pretty good pointers as the first return value is a pandas DataFrame which holds the `ECG_R_Peaks` detector signal (0 for no r peak and 1 for r peak). The second return value contains a dictionary which also holds an `ECG_R_Peaks` array which contains only the sample number at which there was an r peak. This may be easier to use for such an example.

## Thoughts on the contribution itself
Since this `ecg_peaks` function already holds 4 example as of right now, does the new example you created add additional value to them, or would it be better to update one of the examples to also try and show where the r peak values are returned? I.e., I am wondering what the existing confusion is with the function what is best to change.

Thank you again and I look forward to hearing from you.

PR #1153, opened 2025-12-02
[Improvement] Make some ecg_peaks methods accept sample rate as a float

atpage:
A few of the peak detectors in ecg_findpeaks.py (namely `khamis2016`, `zong2003`, `emrich2023`) will fail if sample rate is given as a float, like 1000.0 instead of 1000, due to its use in computing array indices.  This patch converts sample rate to an integer in those cases.  To be safe, the conversion only happens when the value is exact; i.e., a sample rate of 999.99 will still fail.

### DIFF
--- a/neurokit2/ecg/ecg_findpeaks.py
+++ b/neurokit2/ecg/ecg_findpeaks.py
@@ -129,17 +129,6 @@ def _ecg_findpeaks_findmethod(method):
         )
 
 
-def _to_int(f):
-    """Convert f from float to int, if that doesn't require rounding.
-    This is useful when we want to use f to compute integer indices in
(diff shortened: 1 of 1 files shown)


DerAndereJohannes:
Hi atpage, sorry for the late response.

I generally like this idea of having a sophisticated method to do type conversions between different types. I think this function should however be put in `NeuroKit/neurokit2/misc/type_converters.py`. I also like the finite checking recommended by the bot.

If you are still up for making the changes, I would appreciate it a lot!

Thank you again.

DominiqueMakowski:
I reckon having a sophisticated conversion is a bit overkill and adds unnecessary complexity, better to just fix the initial issue with a more lightweight option :) thanks @atpage !

PR #1156, opened 2026-01-09
[Major] Rework the development workflow of Neurokit2

DerAndereJohannes:
> NOTE: this is an ongoing pull request that should be reviewed more thoroughly, but shows a way to make development on Neurokit more maintainable

# Description
One of the main challenges I feel that is present when developing for Neurokit2 is the tooling that we have provided for the developers. If you don't have your IDE setup already, then formatting becomes very manual and not easy to set up. Additionally, better tools exist now which can speedup many repetitive processes while developing and also in github actions itself.

Documentation building is very slow and not trivial to get going. Dependencies and settings are stored all over the repository, in different files and also the github action workflows themselves. Settings are not all in one place.

The main contribution of this PR is to move the development to use the [uv](https://docs.astral.sh/uv/) tool as the core tool of the repository. With one dependency, we can simplify each of the above listed processes and modernize the repository to use the newest standards. It can download its own python interpreters, manage all the dependencies and run each of the tools that are needed for each of the new processes.

Of course, we should still allow legacy developers to use (e.g., using pip) to contribute to. This PR does not make it any more difficult to do so. In fact, it makes it easier for them too.

Overall, with the proposed changes we would be able to onboard people faster and keep the styling better under control. 

# Proposed Changes
I modernized the tools of the following workflows:
1. Merged all the dependency and tool configuration files to the modern `pyproject.toml` standard.
2. Changed the build system to use hatchling over setuptools
3. Moved the main development toolkit to [uv](https://docs.astral.sh/uv/).
4. Moved the main formatting tool from isort, black etc. to [ruff](https://docs.astral.sh/ruff/).
5. Added many different dependency groups such that full neurokit installations (e.g., with all optional dependencies) is supported out of the box.
6. Updated the github workflows such that they no longer contain misc. dependencies and only show overall actions. This includes also fixing workflows that did nothing like `style.yml` and also deduplicating work like in the `documentation builder` workflows (merged to one).
7. Implemented automatic versioning such that it is dealt with via source code versioning rather than strings in a file.
8. Provided a cross-platform `dev.py` that allows developers to quickly run different tasks without needing to install any other dependencies.
9. Added git pre-commits to fix any styling errors or warn the user of invalid styles before it reaches a branch on github.
10. Added a simple flake.nix file for linux/macos users that use nix to manage dependencies (activate with `nix develop`).
11. Parrallelized the sphinx-build such that we can build the documentation much faster. 
12. Python 3.14 support

# What is missing?
Before merge, the documentation for the changes with new updates to the Contribution documentation should be done. Also more work has to be done to make sure that all the Pypi publishing is done correctly as I have reworked it to use [OpenID Connect](https://docs.github.com/en/actions/reference/security/oidc). 

I would appreciate any support and feedback to the PR. Thanks in advance.

### DIFF
--- a/.github/workflows/docs-build.yml
+++ b/.github/workflows/docs-build.yml
@@ -1,64 +1,64 @@
 name: "📖 Documentation"
 
 on:
-  push:
-    branches: [master]
-  pull_request:
-    branches: [master, dev]
+    push:
+        branches: [master]
+    pull_request:
+        branches: [master, dev]
 
 permissions:
-  contents: read
-  pages: write
-  id-token: write
+    contents: read
+    pages: write
+    id-token: write
 
 jobs:
-  build-and-deploy:
-    runs-on: ubuntu-latest
-    env:
-      MNE_DATA: ${{ github.workspace }}/mne_data
-    steps:
-      - name: Checkout Code
-        uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
+    build-and-deploy:
+        runs-on: ubuntu-latest
+        env:
+            MNE_DATA: ${{ github.workspace }}/mne_data
+        steps:
+            - name: Checkout Code
+              uses: actions/checkout@v4
+              with:
+                  fetch-depth: 0
 
-      - name: Install uv
-        uses: astral-sh/setup-uv@v5
-        with:
-          enable-cache: true
-          python-version: "3.13"
+            - name: Install uv
+              uses: astral-sh/setup-uv@v5
+              with:
+                  enable-cache: true
+                  python-version: "3.13"
 
-      - name: Cache MNE Data
-        id: cache-mne
-        uses: actions/cache@v4
-        with:
-          path: ${{ env.MNE_DATA }}
-          key: mne-datasets-v1
+            - name: Cache MNE Data
+              id: cache-mne
+              uses: actions/cache@v4
+              with:
+                  path: ${{ env.MNE_DATA }}
+                  key: mne-datasets-v1
 
-      - name: Install dependencies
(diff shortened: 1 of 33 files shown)


DominiqueMakowski:
Hi @DerAndereJohannes FYI I'll try to get back to NK within the next few days but in the meantime please do as you see fit, and thanks a TON 🙏

DerAndereJohannes:
Thank you for the ongoing encouragement! Take your time to get into the new year :)

I am pretty happy overall with the merge request (and would love to merge it), but there are a few things I would need your input esp. regarding workflows (as you have more permissions and experience than I do):

1. In particular, the release workflows with how it would be uploaded to pypi, I changed it to use OIDC and I am not sure it it would even work :p (Maybe additional setup needed)
2. I am wondering if we still need the `codeql.yml` workflow given we have dependabot and code scanning set via github on the security tab? This would be one more workflow less!
3. I have changed the documentation workflow set to only update when there is a push to master (stable releases). I feel like this is more useful to the majority of users.
4. Additionally, we can look at running the documentation builder using the sphinx-multiversion such that we can have a page for the current stable release and for the new development build (I think this is not much effort to setup).

I look forward to hearing from you!

DominiqueMakowski:
1. I don't really remember this, I think the main thing was to have a pypi and pypi test account, and then the workflow kinda worked, but if it fails I can look into that
2. yes let's remove it, the lighter the better
3. 👌 
4. 👌 multiversion would be cool but whatever is the easiest :)

DerAndereJohannes:
I see you have been extremely productive today! I don't understand your comment, do you want to remove the `cov` tests? Or are you asking about the codecov.yml file?

Issue #869, opened 2023-07-16
Error in basSQI calculation (See ecg_quality.py)?


Abdul099:
Commit: f6d4d2d by @Jacob-Liam-Albaugh 

The formula used to calculate basSQI was updated recently from 1 - (num_power/dem_power) to (1-num_power)/dem_power, where num_power is the psd of the frequencies 0-1Hz and dem_power is the psd of the frequencies 0-40Hz). 

The aim of the SQI is to evaluate the baseline wander in the ECG signal. According to the referenced paper (Zhao 2018), an optimal basSQI lies between 0.95 and 1 whereas a poor basSQI is below 0.95. 

Using the old formula, if we assume an ideal signal has zero psd for 0-1Hz and some value of psd for 0-40Hz, we would get a basSQI of 1-(0/dem_power) = 1. However, with the new formula, we would get a basSQI value of (1-0)/dem_power = 1/dem_power, which could evaluate to a number close to 0 depending how big dem_power is. 

The formula in the referenced paper is in fact the new one :(1-num_power)/dem_power. However, this does not align with the purpose of the index itself.

Thank you in advance!

DominiqueMakowski:
If I understand, you would recommend reverting to the old one?

Abdul099:
@DominiqueMakowski exactly. You can also check this paper where basSQI is calculated as P(1-40)/P(0-40): https://www.sciencedirect.com/science/article/pii/S1746809419302472

DominiqueMakowski:
Then we should probably do that if that makes sense, but document the rationale as a comment in the code + in the docstrings

PR #1164, opened 2026-02-18
Modify basSQI scaling and power ratio calculation

way-ze:
**Describe the bug**
This PR addresses two distinct bugs regarding the `basSQI` calculation in `ecg_quality.py`: a mathematical typo replicating a published error, and a unit mismatch between evaluation modes. This directly resolves the issues raised in #869 regarding `basSQI` returning values > 100.

**1. The Mathematical Typo in `_ecg_quality_basSQI`**
Currently, the function returns `(1 - num_power) / dem_power`. This faithfully replicates a typesetting error printed in Equation 11 of the original [Zhao et al. (2018)](https://www.frontiersin.org/journals/physiology/articles/10.3389/fphys.2018.00727/full) manuscript. 
Proof of the typo exists in the authors' own text immediately below the equation: *"If no baseline drift interference is noted, the basSQI value is close to 1."*
* **Current Code:** `(1 - 0) / 0.005 = 200` (Fails the definition).
* **Corrected Math:** `1 - (0 / 0.005) = 1` (Matches the definition).

**2. The Unit Mismatch in `_ecg_quality_zhao2018`**
The Zhao et al. (2018) methodology shifts units depending on the logic applied. The simple logic thresholds evaluate `basSQI` as a 0-1 decimal. However, the fuzzy logic membership functions (Equations 30-32) scale the baseline variable to a 0-100 percentage. 
Currently, passing the decimal ratio into `mode="fuzzy"` shatters the Cauchy distribution bounds (which check `if basSQI <= 90:`). 

**Proposed Fixes**
1. **Fix the equation in `_ecg_quality_basSQI`:**
Change: `return (1 - num_power) / dem_power`
To: `return 1 - (num_power / dem_power)`

2. **Fix the unit scaling in `_ecg_quality_zhao2018`:**
Add a localized conversion strictly inside the fuzzy block so `mode="simple"` is not affected.
```python
    # UbH
    basSQI = basSQI * 100
    if basSQI <= 90:
        ...

### DIFF
--- a/neurokit2/ecg/ecg_quality.py
+++ b/neurokit2/ecg/ecg_quality.py
@@ -384,7 +384,7 @@ def _ecg_quality_zhao2018(
         if V < 1.5:
             return "Excellent"
         elif V >= 2.40:
-            return "Unnacceptable"
+            return "Unacceptable"
         else:
             return "Barely acceptable"
 


way-ze:
Hi @DerAndereJohannes, thanks for your reply!
I believe this is also a typo in equations 30-32.
image
I believe the boundaries should all be `b` (for example, `90 < b < 95` instead of `90 < q < 95`), the function would not make sense to have `b` as an argument otherwise. The paper has these minor errors elsewhere (for example in equations 26-28, the argument of `U_pH` is `p` but the cases all use `x`as an argument.

way-ze:
Also small typo in line 386 of `ecg_quality`: the label should be "Unacceptable" and not "Unnacceptable"

PR #1166, opened 2026-02-23
[Hotfix] Trigger Pypi update

DominiqueMakowski:
Well it seems like we need to fix some stuff, perhaps we should just re-do a dev PR without increasing the same number? Or we increase by one nonetheless?

### DIFF
--- a/neurokit2/ecg/ecg_quality.py
+++ b/neurokit2/ecg/ecg_quality.py
@@ -235,9 +235,7 @@ def _ecg_quality_averageQRS(ecg_cleaned, rpeaks=None, sampling_rate=1000):
 # =============================================================================
 # Zhao (2018) method
 # =============================================================================
-def _ecg_quality_zhao2018(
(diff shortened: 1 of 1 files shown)


DominiqueMakowski:
also, what should we do with these dependabot PRs? Is there a way to reroute them  to dev?

DerAndereJohannes:
> also, what should we do with these dependabot PRs? Is there a way to reroute them to dev?

I don't have the permissions to update the dependabot settings. Maybe check the settings that are available on it on github.. I will also update the lock file though so we can close all the requests

DerAndereJohannes:
Try and take a look at this link. I don't have access to it, but maybe you can adjust the dependabot settings there

https://github.com/neuropsychology/NeuroKit/settings/dependabot_rules

DominiqueMakowski:
I've disabled the dependabot PRs let me know if there's something else to do

PR #1179, opened 2026-03-02
0.3.0

DominiqueMakowski:
Roadmap

### RSP

### PPG
  - [PyPPG](https://github.com/godamartonaron/GODA_pyPPG)

### Misc
- [ ] Readme looks ugly but this is a global issue (https://github.com/orgs/community/discussions/113792)
- [ ] Complexity : 
  - [ ] Check (and eventually copy) improvements done in [nolds](https://github.com/CSchoel/nolds/pull/41#event-13562169821)
  - [ ] https://www.pnas.org/doi/10.1073/pnas.2420252122


DominiqueMakowski:
@DerAndereJohannes sorry ot bother you aggain, was curious what do you think is the best pipeline & process to autoformat, should we run ruff format manually before comitting?

DerAndereJohannes:
Have you installed pre-commit for your local version of the repository? This should automatically do the formatting on commit! Or did that somehow fail?

Also since you are targeting pandas version 3, this might be something a bit more major, should we consider doing v1.0.0? Then we can start to use the last number as quick patches if e.g., a dependency needs to be capped etc. Like in the example of pandas, we would just do for version v1.0.1 instead of needing to do a v0.2.14 - Which looks a bit more major

DominiqueMakowski:
> Have you installed pre-commit for your local version of the repository? 

How do you do that?

> should we consider doing v1.0.0?

Instead of 0.3.0? you're probably right yes, I think we're long overdue having a 1.0 :)

DerAndereJohannes:
https://neuropsychology.github.io/NeuroKit/resources/contributing.html#setting-up-your-environment
And
https://neuropsychology.github.io/NeuroKit/resources/contributing.html#pre-commit-hooks-automatic-formatting ;-)

Yeah, I think we should respond quite quick to install errors when they happen and I think this is the best way of doing it

DominiqueMakowski:
<sub>*ngl it's a bit embarrassing of me not to have bothered looking into our docs but alas'* 😅 </sub>

DerAndereJohannes:
Happens to the best of us 😉

PR #1181, opened 2026-03-06
Safeguards to avoid crashing when using external R peaks 

Viri1990:
Three checks to avoid crashing when sending external R peaks to delineate the ECG

### DIFF
--- a/neurokit2/ecg/ecg_delineate.py
+++ b/neurokit2/ecg/ecg_delineate.py
@@ -763,7 +763,7 @@ def _calc_prominence(peaks, sig, Rpeak=None, minima=False):
     _sig = -sig if minima else sig
     w[peaks] = scipy.signal.peak_prominences(_sig, peaks)[0]
     # optional: set rpeak prominence to zero to emphasize other peaks
-    if Rpeak is not None:
+    if Rpeak is not None and Rpeak < len(w):
         w[Rpeak] = 0
     return w
 
--- a/neurokit2/ecg/ecg_delineate.py
+++ b/neurokit2/ecg/ecg_delineate.py
@@ -763,7 +763,7 @@ def _calc_prominence(peaks, sig, Rpeak=None, minima=False):
     _sig = -sig if minima else sig
     w[peaks] = scipy.signal.peak_prominences(_sig, peaks)[0]
     # optional: set rpeak prominence to zero to emphasize other peaks
-    if Rpeak is not None and Rpeak < len(w):
+    if Rpeak is not None and 0 <= Rpeak < len(w):
         w[Rpeak] = 0
     return w
 
@@ -781,9 +781,9 @@ def _prominence_find_s_wave(sig, weight_minima, current_wave, max_qrs_interval):
     if "ECG_Q_Peaks" not in current_wave:
         return
     s_bound = current_wave["ECG_Q_Peaks"] + max_qrs_interval
-    s_wave = np.argmax(weight_minima[current_wave["ECG_R_Peaks"] : s_bound] > 0) + current_wave["ECG_R_Peaks"]
     if len(weight_minima[current_wave["ECG_R_Peaks"] : s_bound] > 0) == 0:
         return
+    s_wave = np.argmax(weight_minima[current_wave["ECG_R_Peaks"] : s_bound] > 0) + current_wave["ECG_R_Peaks"]
     current_wave["ECG_S_Peaks"] = (
         np.argmin(sig[current_wave["ECG_R_Peaks"] : s_bound]) + current_wave["ECG_R_Peaks"]
         if s_wave == current_wave["ECG_R_Peaks"]
--- a/neurokit2/ecg/ecg_delineate.py
+++ b/neurokit2/ecg/ecg_delineate.py
@@ -772,7 +772,8 @@ def _prominence_find_q_wave(weight_minima, current_wave, max_r_rise_time):
     if "ECG_R_Peaks" not in current_wave:
         return
     q_bound = max(current_wave["ECG_R_Peaks"] - max_r_rise_time, 0)
-
+    if len(weight_minima[q_bound : current_wave["ECG_R_Peaks"]]) == 0:
+        return
     current_wave["ECG_Q_Peaks"] = np.argmax(weight_minima[q_bound : current_wave["ECG_R_Peaks"]]) + q_bound
 
 
@@ -781,6 +782,8 @@ def _prominence_find_s_wave(sig, weight_minima, current_wave, max_qrs_interval):
         return
     s_bound = current_wave["ECG_Q_Peaks"] + max_qrs_interval
     s_wave = np.argmax(weight_minima[current_wave["ECG_R_Peaks"] : s_bound] > 0) + current_wave["ECG_R_Peaks"]
+    if len(weight_minima[current_wave["ECG_R_Peaks"] : s_bound] > 0) == 0:
+        return
     current_wave["ECG_S_Peaks"] = (
         np.argmin(sig[current_wave["ECG_R_Peaks"] : s_bound]) + current_wave["ECG_R_Peaks"]
         if s_wave == current_wave["ECG_R_Peaks"]


DerAndereJohannes:
Hi Viri1990, thank you for the request.

I see you are trying to skip any r peaks that are found outside the range and trying to account for those. I am wondering if silently skipping them is the optimal choice here, or if we should throw an error at the start of the delineate function if the r peaks indices extend past the data array.

I worry that people might think that everything has run normally, without issue, while maybe their r-peak indices list had an offset of sorts which led to all the r-peaks to actually be in the wrong place etc.

Did this happen to you where an external R peak array extended past the data range? How did that happen?

What do you think?

Thank you!

Viri1990:
> Hi Viri1990, thank you for the request.
> 
> I see you are trying to skip any r peaks that are found outside the range and trying to account for those. I am wondering if silently skipping them is the optimal choice here, or if we should throw an error at the start of the delineate function if the r peaks indices extend past the data array.
> 
> I worry that people might think that everything has run normally, without issue, while maybe their r-peak indices list had an offset of sorts which led to all the r-peaks to actually be in the wrong place etc.
> 
> Did this happen to you where an external R peak array extended past the data range? How did that happen?
> 
> What do you think?
> 
> Thank you!

Hii Johannes, 

The problem that I faced here is that the Rpeaks used to split the ECG signal in beats ("ecg_sig" in line 800) is not taking into account the corner case in which the Rpeak is located at the boundary, so the ecg_sig goes from 0 to n, and the Rpeak is located at n, so the typical IndexError raises and crashes the execution. 
This scenario seems appears in noisy signals because the Rpeak obviously never should be located at the boundary of the beat, so I guess no big problems can arise applying that safety checks. 
Tell me if you need more clarify or wants to apply a different solution.

Thanks!!

DerAndereJohannes:
Hi Viri, sorry for the late response.

I just looked at this section of code for 30 minutes and wondered how reliable it is overall.

Firstly, I think the length checks should be replaced by any checks to see if there are even valid values in the result. If not, then we could return anyway.

However, then I looked at the fallback code that comes afterwards for the s wave and saw that if s == r, then it does a minima check on the raw data itself. This feels pretty hacky and I wonder how reliable this even is. My opinion would be to actually remove this fallback, but I wonder about what you think.

I have posted a diff below of my exploration (which also keeps the minima fallback in, but essentially is dead code) to get your opinion on this. I have also added more comments to the code since it was very unclear without it. It is also more robust then the length check since for the s wave, the length will always be more than 0 since its including the rpeak in the check. I just also changed the q wave to use this too for cleanliness.

Looking forward to your answer .

```diff
diff --git a/neurokit2/ecg/ecg_delineate.py b/neurokit2/ecg/ecg_delineate.py
index 49d919675..45e97c772 100644
--- a/neurokit2/ecg/ecg_delineate.py
+++ b/neurokit2/ecg/ecg_delineate.py
@@ -772,18 +772,31 @@ def _prominence_find_q_wave(weight_minima, current_wave, max_r_rise_time):
     if "ECG_R_Peaks" not in current_wave:
         return
     q_bound = max(current_wave["ECG_R_Peaks"] - max_r_rise_time, 0)
-    if len(weight_minima[q_bound : current_wave["ECG_R_Peaks"]]) == 0:
+
+    # Keep to minima values as we want the deepest trough before the R
+    q_potential = weight_minima[q_bound : current_wave["ECG_R_Peaks"]]
+
+    if not q_potential.any():
         return
-    current_wave["ECG_Q_Peaks"] = np.argmax(weight_minima[q_bound : current_wave["ECG_R_Peaks"]]) + q_bound
+
+    current_wave["ECG_Q_Peaks"] = np.argmax(q_potential) + q_bound
 
 
 def _prominence_find_s_wave(sig, weight_minima, current_wave, max_qrs_interval):
     if "ECG_Q_Peaks" not in current_wave:
         return
     s_bound = current_wave["ECG_Q_Peaks"] + max_qrs_interval
-    if len(weight_minima[current_wave["ECG_R_Peaks"] : s_bound] > 0) == 0:
+
+    # Change to a true list since we want the first downward peak
+    s_potential = weight_minima[current_wave["ECG_R_Peaks"] : s_bound] > 0
+
+    # r peak is 0, so we can keep it in and check if there is an s wave
+    if not (s_potential).any():
         return
-    s_wave = np.argmax(weight_minima[current_wave["ECG_R_Peaks"] : s_bound] > 0) + current_wave["ECG_R_Peaks"]
+
+    s_wave = np.argmax(s_potential) + current_wave["ECG_R_Peaks"]
+
+    # If s is the same as r, check for the lowest value in the qrs bound
     current_wave["ECG_S_Peaks"] = (
         np.argmin(sig[current_wave["ECG_R_Peaks"] : s_bound]) + current_wave["ECG_R_Peaks"]
         if s_wave == current_wave["ECG_R_Peaks"]

```

Viri1990:
> Hi Viri, sorry for the late response.
> 
> I just looked at this section of code for 30 minutes and wondered how reliable it is overall.
> 
> Firstly, I think the length checks should be replaced by any checks to see if there are even valid values in the result. If not, then we could return anyway.
> 
> However, then I looked at the fallback code that comes afterwards for the s wave and saw that if s == r, then it does a minima check on the raw data itself. This feels pretty hacky and I wonder how reliable this even is. My opinion would be to actually remove this fallback, but I wonder about what you think.
> 
> I have posted a diff below of my exploration (which also keeps the minima fallback in, but essentially is dead code) to get your opinion on this. I have also added more comments to the code since it was very unclear without it. It is also more robust then the length check since for the s wave, the length will always be more than 0 since its including the rpeak in the check. I just also changed the q wave to use this too for cleanliness.
> 
> Looking forward to your answer .
> 
> ```diff
> diff --git a/neurokit2/ecg/ecg_delineate.py b/neurokit2/ecg/ecg_delineate.py
> index 49d919675..45e97c772 100644
> --- a/neurokit2/ecg/ecg_delineate.py
> +++ b/neurokit2/ecg/ecg_delineate.py
> @@ -772,18 +772,31 @@ def _prominence_find_q_wave(weight_minima, current_wave, max_r_rise_time):
>      if "ECG_R_Peaks" not in current_wave:
>          return
>      q_bound = max(current_wave["ECG_R_Peaks"] - max_r_rise_time, 0)
> -    if len(weight_minima[q_bound : current_wave["ECG_R_Peaks"]]) == 0:
> +
> +    # Keep to minima values as we want the deepest trough before the R
> +    q_potential = weight_minima[q_bound : current_wave["ECG_R_Peaks"]]
> +
> +    if not q_potential.any():
>          return
> -    current_wave["ECG_Q_Peaks"] = np.argmax(weight_minima[q_bound : current_wave["ECG_R_Peaks"]]) + q_bound
> +
> +    current_wave["ECG_Q_Peaks"] = np.argmax(q_potential) + q_bound
>  
>  
>  def _prominence_find_s_wave(sig, weight_minima, current_wave, max_qrs_interval):
>      if "ECG_Q_Peaks" not in current_wave:
>          return
>      s_bound = current_wave["ECG_Q_Peaks"] + max_qrs_interval
> -    if len(weight_minima[current_wave["ECG_R_Peaks"] : s_bound] > 0) == 0:
> +
> +    # Change to a true list since we want the first downward peak
> +    s_potential = weight_minima[current_wave["ECG_R_Peaks"] : s_bound] > 0
> +
> +    # r peak is 0, so we can keep it in and check if there is an s wave
> +    if not (s_potential).any():
>          return
> -    s_wave = np.argmax(weight_minima[current_wave["ECG_R_Peaks"] : s_bound] > 0) + current_wave["ECG_R_Peaks"]
> +
> +    s_wave = np.argmax(s_potential) + current_wave["ECG_R_Peaks"]
> +
> +    # If s is the same as r, check for the lowest value in the qrs bound
>      current_wave["ECG_S_Peaks"] = (
>          np.argmin(sig[current_wave["ECG_R_Peaks"] : s_bound]) + current_wave["ECG_R_Peaks"]
>          if s_wave == current_wave["ECG_R_Peaks"]
> ```

Hello again Johannes,
Sorry for my late response, I just tested the code and works fine, so if you want to proceed with the merge it'd great!

Thanks for your effort!

PR #1184, opened 2026-03-17
[Fix] Read acqknowledge now tries to convert sample rate to int

### LINKED ISSUE #1182: NK read_acknowledge appears to return a float for signal rate while NK functions expect int input
**Describe the bug**
When calling eda_intervalrelated, signal_autocor is called expecting lag as an int for use as indexing, derived from the provided sample rate. Python does not appear to dynamically coerce to an int in this case, despite my lag value being a whole number (8000.0). Upon investigation, sample rate returned by read_ackowledge appears to be a float, while typing hints from eda_intervalrelated suggest it should be an int.

**To Reproduce**
Read in an acknowledge file
test the type of sample_rate using type()
To verify, i used the following code
```py
import neurokit2 as nk
import os

for file in os.listdir("data"):
    dataFile = "data\\" + file
    for z in os.listdir(dataFile):
        if z.endswith(".csv"):
            continue  # skip csv files, only process acqknowledge files
        if z.endswith(".acq"):
            file = f"data\\{file}\\{z}"
        data, sampling_rate = nk.read_acqknowledge(file)
        print(f"type: {type(sampling_rate)}, value: {sampling_rate}")
```
in a file structure of 
```
src
├test.py
└data
  ├folder1
    └1.acq
  ├folder2
     └2.acq
```
etc. extra code to remove csvs is due to testing in a directory with csv folders. In my dataset of 4 acq files it returned 
```
type: <class 'numpy.float64'>, value: 2000.0
type: <class 'numpy.float64'>, value: 2000.0
type: <class 'numpy.float64'>, value: 2000.0
type: <class 'numpy.float64'>, value: 2000.0
```
 I would expect 4 ints.

**Expected behaviour**
If nk functions expect signal rate to be an int read_ackowledge should be returning an int without requiring additional coercion from the user.

**Screenshots**
If applicable and useful, you can add screenshots by drag-and-dropping the images in the issue window.

**System Specifications**

It's important that you give us some information about the system you are using. For that you can run:

```python
import neurokit2 as nk
nk.version()
```

- OS: Windows (WindowsPE 64bit) 
- Python: 3.11.4 
- NeuroKit2: 0.2.12 
- NumPy: 2.4.2 
- Pandas: 2.3.3 
- SciPy: 1.17.0 
- sklearn: 1.8.0 
- matplotlib: 3.10.8

Lyroth001:
# Description
Ensures that read_acqknowledge() returns sampling rate as a python native int, not a float. Also updates read_acqknowledge() unit test to check for typing

fixes #1182 

# Proposed Changes

Adds automated type coercion to int in read_acqknowledge
Adds a warning if this coercion fails
Updates the test for this function to check the type output

# Checklist

- [x] I have added the newly added features to **News.rst** (if applicable) -na i think

### DIFF
--- a/neurokit2/data/read_acqknowledge.py
+++ b/neurokit2/data/read_acqknowledge.py
@@ -8,6 +8,7 @@ import pandas as pd
 from ..misc import NeuroKitWarning
 from ..signal import signal_resample
 
+
 def read_acqknowledge(filename, sampling_rate="max", resample_method="interpolation", impute_missing=True):
     """**Read and format a BIOPAC's AcqKnowledge file into a pandas' dataframe**
 
--- a/neurokit2/data/read_acqknowledge.py
+++ b/neurokit2/data/read_acqknowledge.py
@@ -5,8 +5,8 @@ from warnings import warn
 import numpy as np
 import pandas as pd
 
-from ..signal import signal_resample
 from ..misc import NeuroKitWarning
+from ..signal import signal_resample
 
 def read_acqknowledge(filename, sampling_rate="max", resample_method="interpolation", impute_missing=True):
     """**Read and format a BIOPAC's AcqKnowledge file into a pandas' dataframe**
--- a/neurokit2/data/read_acqknowledge.py
+++ b/neurokit2/data/read_acqknowledge.py
@@ -84,9 +84,9 @@ def read_acqknowledge(filename, sampling_rate="max", resample_method="interpolat
         else:
             sampling_rate = max_freq
             warn(
-                f"Automatic sampling rate conversion to int failed"
-                f"Sampling rate is {sampling_rate}, of type {type(sampling_rate)}"
-                f"This may cause errors or unexpected behaviours of other neurokit functions",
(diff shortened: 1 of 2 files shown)


DerAndereJohannes:
Hey, nice job! Thank you a lot for the tests too. They are always greatly appreciated. I am sure the nan value will never appear in the wild, but I am glad to have the check and the warning.

Regarding the style formatting: It was just an extra line added after the imports. I see how the error was a bit confusing. I probably should have mentioned that we have a contributing guide that shows the tools that we use to make sure the formatting is consistent in every file [here](https://neuropsychology.github.io/NeuroKit/resources/contributing.html#setting-up-your-environment). Usually, when you have the pre-commits installed, it would have automatically fixed the formatting errors for you.

Let me know if you have any other questions, I will merge after the checks have completed.

Thanks!

Lyroth001:
Ahh i assumed it wanted alphabetical or something lol

And yeah, agree that its always better to cover a weird case you never expect to see than have it silently occur :p

PR #1186, opened 2026-03-22
[Feature] CWT delineation detects inverted T and P waves

### LINKED ISSUE #1138: Misdetections of `ecg_delineate()` with CWT on signals with T-wave inversions
Hi!

I wanted to perform an ML pipeline for cardiac diseases using hand-crafted features computed from the output of `ecg_delineate()`, so I need a delineation that works with ECGs from patients with a disease.
 
- Do all the algorithms operate under the assumption that R-waves and T-waves are concave? 

From prior discussions, I learned that if the signal is inverted, `ecg_invert()` should be applied. However, there are some scenarios where despite having concave P-waves and R-waves, the T-wave is convex (i.e. inverted). 

### Example
Here you have an example from [Lobachevsky University Electrocardiography Database](https://physionet.org/content/ludb/1.0.1/). In the lead II of the record 11, the ground truth T-wave is inverted. However, when applying `ecg_delineate()` using CWT it detects a maxima as the ECG_T_Peak (red line):

Image

DerAndereJohannes:
Massively improved inverted T and P waves at a small cost for T peak detection in upright data. (93.7% recall vs 91.1% recall in whole LUDB dataset).

Here is an example of the output as initially asked for in issue #1138:
 
ludb_r11_ii

It is not perfect though, take a look at record 4 with the avr lead (you can see the R onset/offset are a bit off):

ludb_r4_avr

Difficult situations still remain like for record 11 avr lead which basically has a flatline of a T wave:

ludb_r11_avr

# Description

The current implementation just checks for negative to positive zero-crossings for the T (first detected peak after R) and P (last detected peak after R) wave.

# Proposed Changes

The new method recognises both crossing directions and keeps an array on each peak as upright or inverted. T and P are selected based on maximum amplitude rather than the temporal place in the current solution. T is checked 80-400ms after R. P is a bit more complicated as it had mixed results. The change I add suggests checking for the largest upright candidate in the 50-300ms windows before the next R, but falls back to the largest inverted candidate if none exists.

The polarity values are then passed onto the onset/offset as normal. 

# Asking the community

This is quite a major change as to how the `ecg_delineate` cwt method works and . I would like to ask for people to test out these changes first and gather feedback before merging these changes into dev. I would not want to make the most common data type (Lead II with upright p and t waves) to suffer for the changes. Putting this in a separate method would also be an option. Thanks!

closes #1138

### DIFF
--- a/neurokit2/ecg/ecg_delineate.py
+++ b/neurokit2/ecg/ecg_delineate.py
@@ -645,16 +645,20 @@ def _dwt_compute_multiscales(ecg: np.ndarray, max_degree):
 # =============================================================================
 def _ecg_delineator_cwt(ecg, rpeaks=None, sampling_rate=1000):
     # P-Peaks and T-Peaks
-    tpeaks, ppeaks = _peaks_delineator(ecg, rpeaks, sampling_rate=sampling_rate)
+    tpeaks, ppeaks, t_polarities, p_polarities = _peaks_delineator(ecg, rpeaks, sampling_rate=sampling_rate)
 
     # qrs onsets and offsets
     qrs_onsets, qrs_offsets = _onset_offset_delineator(ecg, rpeaks, peak_type="rpeaks", sampling_rate=sampling_rate)
 
     # ppeaks onsets and offsets
-    p_onsets, p_offsets = _onset_offset_delineator(ecg, ppeaks, peak_type="ppeaks", sampling_rate=sampling_rate)
+    p_onsets, p_offsets = _onset_offset_delineator(
+        ecg, ppeaks, peak_type="ppeaks", sampling_rate=sampling_rate, polarities=p_polarities
+    )
 
     # tpeaks onsets and offsets
-    t_onsets, t_offsets = _onset_offset_delineator(ecg, tpeaks, peak_type="tpeaks", sampling_rate=sampling_rate)
+    t_onsets, t_offsets = _onset_offset_delineator(
+        ecg, tpeaks, peak_type="tpeaks", sampling_rate=sampling_rate, polarities=t_polarities
+    )
 
     # No dwt defined method for Q and S peak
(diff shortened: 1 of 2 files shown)


atpage:
> test out these changes first and gather feedback before merging these changes

I have not tried this new code for inverted P or T waves, but in my own testing, I found that AUC (on a baseline-corrected signal) generally works better than amplitude for determining the directions of these waves.  I also check the entire signal for a consistent direction, rather than letting it flip between upright and inverted on each beat, though this could be a problem if morphology is actually changing with time.

> you can see the R onset/offset are a bit off

In the dataset I'm currently working on, I found QRS onset/offset annotations from neurokit to be completely unreliable.  This is mainly because neurokit seems to expect all 3 waves (Q, R, S), but in reality most leads will have only two waves (i.e. QR or RS).  In those cases, neurokit will "invent" the missing third wave.

My best solution so far is to trust the annotations of only the largest negative wave, which will be either Q or S.  Then I invert the lead and reprocess it, again trusting the annotations of the largest negative wave, which will be either S or Q.  Combining these gives good QRS bounds in the case of QR or RS complexes.  However, when a full QRS complex *is* present, some heuristics are necessary to revert to the original annotations.  I also still need a solution when there are just R waves with no Q or S.

DerAndereJohannes:
Hi Atpage, thank you for taking the time to look into it. And sorry I am just now getting back to you.
> This is mainly because neurokit seems to expect all 3 waves (Q, R, S)...

From me working on this pull request, I can tell you that this is correct and I see how this is an issue, especially when working with other leads. It looks like there is a lot more work to be done for your type of solution to be implemented.

I might try and play around with the pseudocode that you provided in the other issue, though I cannot guarantee that I will have time to implement this at the current moment.

I will close this pull request for now as insufficient. Thank you for your time!

